PR #1, opened 2024-05-30
Add node creation menu

wnbaum:
Work in progress, starting to organize how the user can edit nodes. Turned off gdlint temporarily because it wasn't letting me commit with lines longer than 100 which is a little too short for me :P

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -13,7 +13,6 @@ repos:
   # <https://github.com/Scony/godot-gdscript-toolkit/commit/e3c37fa54fb32ad99b0bceb2f17e7bc1ffd99806>
   rev: f836958a4487e31e3c5ab35c57c7a2128b7e2303
   hooks:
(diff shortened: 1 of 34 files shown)


wnbaum:
So, the idea is that the user first creates a node that can inherit from presets we have made, for example a `SimpleCharacter` with a Sprite and a collision box. Then, they would click/select the new node, and add a block script, essentially extending the functionality of the pre-made scene. Finally, they could instantiate their node and add it to the scene.

Maybe we could add an advanced option later to expose core Godot nodes to this menu. Or another advanced option to enable nesting in the scene.

Next step is to persist block script changes when switching between created nodes.

wnbaum:
Just added another commit with node picker categories since it may conflict with changes to the script editor windows.

wnbaum:
Fixed up some stuff, and deprecated the custom node editor for now. We can bring it back in the future but for now are focusing on being able to attach block scripts to nodes.

PR #2, opened 2024-05-31
Persist canvas

manuq:


wnbaum:
Hm. With `PackedScene` we shouldn't need to recursively go through each child node. I did read in the documentation under the `pack(node)` method that it ignores any nodes that are not owned by the given `node`. I tried dynamically setting the owner of new blocks that are added to the window itself, but that didn't seem to work. Will update on Monday.

wnbaum:
Moved to `node-create` branch and merged WIP changes to main.

PR #3, opened 2024-06-04
Add block scripts to nodes

wnbaum:
This PR is for essentially replacing the existing scripting interface with a block scripting interface.

So far I have replaced the "create script" menu, and worked on persisting the nodes using a tree resource with `PackedScene`'s as the data. This seems to be the best option I have found as the child structure of the blocks is complex in that the `PackedScene.pack()` is not well suited for this problem. (To use `pack()` you would need to set the owner of every child node to the root, which apparently makes them get instantiated multiple times.)

Since each block is a scene, we only want to instantiate them, not their children. So the solution I came up with is to store a tree structure where each node is composed of the packed block scene, and paths to their child blocks. This way we can go through the tree on load and only instantiate each packed scene once.

There seems to be some errors related to signals that we need to figure out as well.

https://phabricator.endlessm.com/T35447


manuq:
@wnbaum I wasn't able to attach 2 different block codes to 2 nodes. This is what I tried:
1. From the Scene dock, add 2 Node2D nodes. They get named "Node2D" and "Node2D2" 
1. Right-click on "Node2D" -> Attach Script. The usual dialog appears and it says "GDScript" and the default location is "res://test_game/Node2D.gd". Anyways it seems to work because it opens the Block Code editor and I can see it also creates a "res://test_game/Node2D_bsd.tres" file. The script icon doesn't appear next to the node name.
1. In the Block Code editor, do something for this node like "on ready: print one": 
Captura desde 2024-06-04 10-41-13
1. Right-click on "Node2D2" -> Attach Script. This again creates 2 files: "res://test_game/Node2D2.gd" and "res://test_game/Node2D2_bsd.tres". But the block canvas still shows the blocks corresponding to the first Node2D: 
Captura desde 2024-06-04 10-44-49

manuq:
Converting to draft because this is work in progress.

wnbaum:
Ok, you can now attach scripts like you normally would, but it will open the block script interface instead. You can also click the open script button in the scene tree to open an attached block script.

When moved, block scripts will automatically save. The generated script is stored in the `.gd` file and the block scene data is stored in the `_bsd.tres` (BlockScriptData) file.

In addition, if the script inherits from a script that has the `get_custom_blocks()` method, the picker will be populated with these blocks. It is not super easy to instantiate and extend these scripts yet, (e.g SimpleCharacter) and we can possibly change the UI.

There are some other UI problems that need to be addressed (see README), but for now it is pretty functional so I will merge into main, as I want to work on making the `Block` framework/inheritance tree nicer to work with.

manuq:
This is looking very promissing! I'm happy with the addition of custom resources for persisting.

A couple issues found:
- The generated GDScript is being cached and doesn't end saved in the file, even when the output says "Saved generated script to res://test_game/MyNode.gd" only the `_bsd.tres` file is saved.
- If I create a Node2D it gets named "Node2D", then I attach blocks: I end up with `class_name Node2D extends Node2D` in the generated GDScript which fails with "Parse Error: Class "Node2D" hides a native class."
- Can't double click in a node to rename it anymore.

PR #4, opened 2024-06-05
Refactor block scenes

wnbaum:
Cleaned up old block scenes that were a mix between scene inheritance and properties. Now it is easy to create and format new statement and parameter blocks.
Added a test scene where there is a node that inherits from `SimpleCharacter` called `MyCharacter` that can move around in the scene.

Note: when adding a script to the character, you should 1: save it in a different location and 2: make sure it inherits from `SimpleCharacter` rather than the base node. This is an unintuitive process and we should look at making it simpler.

https://phabricator.endlessm.com/T35448


wnbaum:
An example of creating a block is in `simple_character.gd`. You can format the blocks by supplying parameters with types, and use those types in the generated code statement. This is a custom block associated with the `SimpleCharacter` class. To add global blocks you would write something similar in `category_factory.gd`.
```
b = CategoryFactory.BLOCKS["statement_block"].instantiate()
b.block_type = Types.BlockType.EXECUTE
b.block_format = "Add movement input {input: VECTOR2} with speed {speed: INT}"
b.statement = "velocity = {input}*{speed}\nmove_and_slide()"
```

wnbaum:
Also, a better "control" block can be added next so we can programmatically create blocks such as if/else and loops.

wnbaum:
Merged new way to attach block code scripts. Will continue working on making creating blocks easier.

PR #6, opened 2024-06-06
Add new block code node

wnbaum:
The `BlockCode` node can be attached to any node to give it a block script. Just select it after instancing and open the script by clicking the inspector button. When the script is now edited, it will update the corresponding `BlockScriptData` that is associated with the scene. This way there are no external files stored, and saving the scene will save all of the block scripts. The scene will also save on run which is nice.

Check out the `test_game.tscn` to see a demo.


addons/block_code/block_code_node/block_code.tscn
dylanmccall, line 1: Adding the scene file here means the list of nodes one can add to a project will include both BlockCode the PackedScene and BlockCode the Node (from the gdscript file). I vote for removing the scene file, since it doesn't instantiate anything ;)

dylanmccall (changes requested):

addons/block_code/block_code_node/block_code.tscn
wnbaum: Yes, definitely a good idea! I also made it so when you activate the plugin, the internal plugin classes are removed from the create node popup so that users don't try to use them.

PR #7, opened 2024-06-07
Block code node manuq

manuq:


dbnicholson:
Is this still relevant?

manuq:
I still need to port parts of this to new PRs. Like the way the block script resource emits a signal when changed.

PR #8, opened 2024-06-07
Resource improvements

manuq:


wnbaum:
Looks good! I just wouldn't want to expose the script resource to beginners since it's only really useful for debugging. We can keep it exposed for now though as we are developing. Not sure if we should comment out the `_get_property_list()` method or if it just stay in the git history. Also I don't know what the autoformatter did to it xD. If you think its good as is go ahead and merge it!

addons/block_code/block_code_node/block_code.gd
```diff
@@ -36,22 +29,18 @@ func _enter_tree():
 		plugin.add_inspector_plugin(load("res://addons/block_code/inspector_plugin/block_script_inspector.gd").new())
 
 
-# Necessary to "export" the block script data without exposing it
-func _get_property_list():
```
wnbaum, line 40: Yea, maybe just comment out this function instead of removing it if you want, up to you. Not sure what the standard is.

wnbaum (approved):

addons/block_code/block_code_node/block_code.gd
manuq: Actually it's a bit hard because it involves setting `PROPERTY_USAGE_EDITOR` or `PROPERTY_USAGE_NO_EDITOR` instead of the usual `@export`. I don't think we should hide this, at least for the time being.

PR #9, opened 2024-06-07
New blocks: Signals, physics, and movement

wnbaum:
This adds new blocks for signals. A node can broadcast a signal and any other node can receive it by name. Maybe can pass data through signals too. Added a physics process entry block, and simplified the movement block.

These are all steps towards making a simple game using blocks.

### DIFF
--- a/addons/block_code/block_code_node/block_code.gd
+++ b/addons/block_code/block_code_node/block_code.gd
@@ -14,9 +14,9 @@ func _ready():
 	var script := GDScript.new()
 	script.set_source_code(bsd.generated_script)
 	script.reload()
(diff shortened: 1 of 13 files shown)


manuq:
Early review while this is in draft: It's cool enough for merging! Ultimately the way of communicating nodes with blocks will be instructed by the mockups.

wnbaum:
Added some simple get/set variable blocks as well as integer math. When we iterate more we can decide how to handle types, but for now this will push forward making a simple game. This PR is getting quite big so I would rather merge it and make a new one for the physics area stuff!

manuq:
Excellent!

PR #10, opened 2024-06-10
Add knobs to statement blocks

manuq:
Rather rudimentary for now:
Captura desde 2024-06-10 15-46-42

### DIFF
--- a/addons/block_code/ui/blocks/statement_block/background.gd
+++ b/addons/block_code/ui/blocks/statement_block/background.gd
@@ -1,0 +1,51 @@
+@tool
+extends Control
+
+const KNOB_X = 30.0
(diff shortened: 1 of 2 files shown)


wnbaum (approved):
Super cool, love it! Maybe the cut out and extruded bit can be swapped from top to bottom to look more like blocks that students are used to but it's unique and can obviously be iterated on if we want. Looks good to merge.

manuq:
> Maybe the cut out and extruded bit can be swapped from top to bottom to look more like blocks that students are used to but it's unique and can obviously be iterated on if we want.

Ah true, how you say is also how Blockly or MS Arcade do it. I'll swap the cutouts then.

PR #11, opened 2024-06-10
Blocks UI: Improve knobs in statement blocks

manuq:
https://phabricator.endlessm.com/T35477


wnbaum (approved):
New blocks look super clean, nice!! I updated the basic block to use the new background script, and made it so you can toggle having the top cutout. Also the preview highlight now looks like a block :)

PR #12, opened 2024-06-11
Block Code Node: Add icon

manuq:
https://phabricator.endlessm.com/T35477


dylanmccall:
I like it! And I like the distinct colour choice :)

PR #13, opened 2024-06-11
Control and entry blocks with parameters

wnbaum:
Made it much easier to create new control blocks (if/else, loops), and styled them better.

Also added a way to create parameters in entry blocks, for example you could have an entry point if a body enters an Area2D, and the parameter would be the body entered (see new block in "test" category). Then maybe you could check the name of the body to increase the score or something. Conditional stuff will come next.

When implementing the "control" blocks, I did feel like generating a text script is a little messy, and would probably be cleaner with interpreted blocks. Maybe that is something to explore further, but for now I think this is working fine for a prototype, and we also want to possibly have a readable script for the user to switch to.


manuq:
Looks good! We ultimately would like also an IF without an ELSE, no? MS Arcade has a nice expand feature to convert an IF to an IF/ELSE.

PR #14, opened 2024-06-11
Added group blocks

wnbaum:
Added blocks for node/group stuff and changed existing signal block to signal to a group instead of all nodes. We can change the name to "call" instead of "signal" if we want, but since the user doesn't actually implement their own functions yet, it didn't really make sense to change "on signal" to "on call".

I also made the styling a bit better and added an `if` block without the `else`. You can see an example of the new stuff in the `Manuel` block script lol. Next thing to add can be a simple `Area2D` node that actually generates this `on_body_enter` event.

screenshot


manuq:
Impressive!

PR #17, opened 2024-06-12
Blocks for built-ins

wnbaum:
This adds a way to add custom blocks to a specific built-in class. You can either expose a property (given it's `Variant.Type` can match a `Types.BlockType`), or create a custom block for more control. See `get_built_in_categories()` in `CategoryFactory`.

You can find an example of this for `Node2D` in test scene.

Something to think about at some point is how we can define our types better. Should we have a `NUMBER` type that allows for integer and float input? And do we need a way to have type inheritance, such that we could keep the `INT` and `FLOAT` types, but they are still a `NUMBER`? I think defining our own types is a fine thing to do, as we want to add a layer of abstraction.

I also made the styling a bit nicer, LMK if you don't like it. Essentially just gave the blocks a bit more margin space and gave outlines to parameter blocks.

### DIFF
--- a/addons/block_code/ui/picker/categories/category_factory.gd
+++ b/addons/block_code/ui/picker/categories/category_factory.gd
@@ -192,3 +192,93 @@ static func add_to_categories(main: Array[BlockCategory], addition: Array[BlockC
 			main.append(add_category)
 
 	return main
+
+
+static func built_in_type_to_block_type(type: Variant.Type):
+	match type:
+		TYPE_BOOL:
+			return Types.BlockType.BOOL
+		TYPE_INT:
+			return Types.BlockType.INT
+		TYPE_FLOAT:
+			return Types.BlockType.FLOAT
+		TYPE_STRING:
+			return Types.BlockType.STRING
+		TYPE_VECTOR2:
+			return Types.BlockType.VECTOR2
+
+	return null
+
+
+static func property_to_blocklist(property: Dictionary) -> Array[Block]:
+	var block_list: Array[Block] = []
+
+	var block_type = built_in_type_to_block_type(property.type)
+
+	if block_type:
+		var type_string: String = Types.BlockType.find_key(block_type)
+
+		var b = BLOCKS["statement_block"].instantiate()
+		b.block_format = "Set %s {value: %s}" % [property.name, type_string]
+		b.statement = "%s = {value}" % property.name
+		block_list.append(b)
+
+		b = BLOCKS["parameter_block"].instantiate()
+		b.block_type = block_type
+		b.block_format = "%s" % property.name
+		b.statement = "%s" % property.name
+		block_list.append(b)
+
+	return block_list
+
+
+static func category_from_property_list(property_list: Array, selected_props: Array, p_name: String, p_color: Color) -> BlockCategory:
+	var block_list: Array[Block]
+
+	for selected_property in selected_props:
+		var found_prop
+		for prop in property_list:
+			if selected_property == prop.name:
+				found_prop = prop
+				break
+		block_list.append_array(property_to_blocklist(found_prop))
+
+	return BlockCategory.new(p_name, block_list, p_color)
+
+
+static func get_inherited_categories(_class_name: String) -> Array[BlockCategory]:
+	var cats: Array[BlockCategory] = []
+
+	var current: String = _class_name
+
+	while current != "":
+		add_to_categories(cats, get_built_in_categories(current))
+		current = ClassDB.get_parent_class(current)
+
+	return cats
+
+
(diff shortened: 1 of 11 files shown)


addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -192,3 +192,93 @@ static func add_to_categories(main: Array[BlockCategory], addition: Array[BlockC
...
+		TYPE_BOOL:
+			return Types.BlockType.BOOL
+		TYPE_INT:
+			return Types.BlockType.INT
+		TYPE_FLOAT:
+			return Types.BlockType.FLOAT
+		TYPE_STRING:
+			return Types.BlockType.STRING
+		TYPE_VECTOR2:
+			return Types.BlockType.VECTOR2
```
manuq, line 208: I think there shouldn't be block types that map to Variant types, we can use Variant types directly (see my spec). All these blocks can be a single Variant block that has Variant type as property. But good to keep this for now!

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -192,3 +192,93 @@ static func add_to_categories(main: Array[BlockCategory], addition: Array[BlockC
...
+
+	return null
+
+
+static func property_to_blocklist(property: Dictionary) -> Array[Block]:
```
manuq, line 213: Nice! I love this factory function. It can also have a "Change property by [amount]" that translates to `property += amount` as a sugar syntax to "Set property to [value]" and a "a + b" block. See MS Arcade for reference.

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -192,3 +192,93 @@ static func add_to_categories(main: Array[BlockCategory], addition: Array[BlockC
...
+	var block_list: Array[Block] = []
+
+	var block_type = built_in_type_to_block_type(property.type)
+
+	if block_type:
+		var type_string: String = Types.BlockType.find_key(block_type)
+
+		var b = BLOCKS["statement_block"].instantiate()
+		b.block_format = "Set %s {value: %s}" % [property.name, type_string]
```
manuq, line 222:
```suggestion
		b.block_format = "Set %s to {value: %s}" % [property.name, type_string]
```

Also it would be nice to [capitalize](https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-capitalize) `property.name` like the Inspector does.

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -192,3 +192,93 @@ static func add_to_categories(main: Array[BlockCategory], addition: Array[BlockC
...
+	var block_list: Array[Block] = []
+
+	match _class_name:
+		"Node2D":
+			var b = BLOCKS["statement_block"].instantiate()
+			b.block_format = "Set rotation degrees {angle: FLOAT}"
+			b.statement = "rotation = deg_to_rad({angle})"
+			block_list.append(b)
+
+			props = ["position", "rotation"]
```
manuq, line 274:
```suggestion
			props = ["position", "rotation", "scale"]
```

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -192,3 +192,93 @@ static func add_to_categories(main: Array[BlockCategory], addition: Array[BlockC
...
+	var cats: Array[BlockCategory] = []
+
+	var props: Array = []
+	var block_list: Array[Block] = []
+
+	match _class_name:
+		"Node2D":
+			var b = BLOCKS["statement_block"].instantiate()
+			b.block_format = "Set rotation degrees {angle: FLOAT}"
+			b.statement = "rotation = deg_to_rad({angle})"
```
manuq, line 271: No need to use deg_to_rad because Node2D already has a [rotation_degrees](https://docs.godotengine.org/en/stable/classes/class_node2d.html#class-node2d-property-rotation-degrees) property:
```suggestion
			b.statement = "rotation_degrees = {angle}"
```

manuq (changes requested):
Fantastic work, I left some suggestions.

addons/block_code/ui/picker/categories/category_factory.gd
wnbaum: I agree, the current typing system needs some work, but it is useful to have the BlockType stay as it is now, since there isn't really a type for `EXECUTE` in `Variant.Type`.

addons/block_code/ui/picker/categories/category_factory.gd
wnbaum: This is a great idea and I will add the simple version of it now that adds two of the same type together. I think it will be even better once we are able to add a float to a vector and so on.

PR #19, opened 2024-06-13
Prepare addon for Asset Library

manuq:
Following the submission docs [1]

- Use the export-ignore Git attribute so only the addon is in the resulting ZIP file.

- Copy LICENSE and README to the addon.

- Create an empty docs/screenshot folder for the Asset Library assets like icon, screenshots. And make the Godot editor ignores these files (ie. doesn't automatically import them).

[1] https://docs.godotengine.org/en/latest/community/asset_library/submitting_to_assetlib.html

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,10 @@
 # Normalize EOL for all files that Git considers text files.
 * text=auto eol=lf
+
+# Only include the block coding addon when downloading from the Asset Library.
+# Exclude non-addon files and the plugin_refresher addon used during
+# development.
+/**                    export-ignore
+/addons               !export-ignore
+/addons/block_code    !export-ignore
+/addons/block_code/** !export-ignore


.gitattributes
```diff
@@ -1,2 +1,7 @@
 # Normalize EOL for all files that Git considers text files.
 * text=auto eol=lf
+
+# Only include the addons folder when downloading from the Asset Library.
+/**                   export-ignore
+/addons/block_code    !export-ignore
+/addons/block_code/** !export-ignore
```
wjt, line 7: This actually causes no files to be included. You can test it with:

```
git archive --worktree-attributes --format tar HEAD | tar -t
```

`--worktree-attributes` means that uncommitted changes to `.gitattributes` will be used; it's just for easier development.


I'll push an alternative commit.

wjt (approved):
I hope you don't mind that I pushed new commits rather than describing the changes I suggested. They are:

- Making `.gitattributes` wordier so that it correctly still includes the addon, rather than making `git archive` export an empty ZIP file
- Using a symbolic link rather than copying README.md.

.gitattributes
```diff
@@ -1,2 +1,12 @@
...
+
+# Exclude everything outside /addons/block_code
+/.*                       export-ignore
+/LICENSE                  export-ignore
+/README.md                export-ignore
+/addons/plugin_refresher  export-ignore
+/docs                     export-ignore
+/icon.svg*                export-ignore
+/project.godot            export-ignore
+/test_game                export-ignore
```
wjt, line 12: I re-read the documentation and I think this can be made simpler actually. 1 moment.

wjt: Compared to your original approach there is just 1 line change:

```diff
-# Only include the addons folder when downloading from the Asset Library.
+# Only include the block coding addon when downloading from the Asset Library.
+# Exclude non-addon files and the plugin_refresher addon used during
+# development.
 /**                    export-ignore
+/addons               !export-ignore
 /addons/block_code    !export-ignore
 /addons/block_code/** !export-ignore
```

manuq: Yes, go ahead! Thanks for letting me know how to test this locally.

wjt:
Hmm I just figure out that, because Godot Asset Library uses `git archive --format zip` and the zip format doesn't support symbolic links, we can actually avoid duplicating the license too. I'll open a separate PR.

wjt:
No, it doesn't work, I was just reading the output wrong. Never mind.

manuq:
> No, it doesn't work, I was just reading the output wrong. Never mind.

That was my initial hope as well. It will be messy to keep the files in sync, but hopefully that can be added to the release automation?

PR #20, opened 2024-06-13
Run pre-commit checks in GitHub Actions

wjt:


wnbaum (approved):
Cool! Looks good to me. At some point we should figure out how to change the max line length on [godot-gdscript-toolkit](https://github.com/Scony/godot-gdscript-toolkit)'s gdlint. There was an option in gdformat, but not gdlint... After not being able to commit for an hour because my paths were too long I ended up just removing gdlint lol. Maybe I am just silly for wanting longer line lengths though?

PR #21, opened 2024-06-13
github: Add workflow to publish to Asset Library

wjt:
The workflow is triggered by publishing a release (which is essentially
a fancy tag). If it turns out we need to trigger it manually we could
later add workflow_dispatch.

### DIFF
--- a/.github/workflows/godot-asset-library.yaml
+++ b/.github/workflows/godot-asset-library.yaml
@@ -1,0 +1,21 @@
+on:
+  release:
+    types:
+      - published
+
+name: Push to Godot Asset Library
+
+jobs:
+  publish:
+    if: github.event.action == 'released'
+    runs-on: ubuntu-latest
+    name: Push new release
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Push to Godot Asset Library
+        uses: deep-entertainment/godot-asset-lib-action@v0.4.0
+        with:
+          username: ${{ secrets.GODOT_ASSET_LIBRARY_USERNAME }}
+          password: ${{ secrets.GODOT_ASSET_LIBRARY_PASSWORD }}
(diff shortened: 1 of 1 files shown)


.github/workflows/godot-asset-library.yaml
```diff
@@ -0,0 +1,21 @@
...
+    name: Push new release
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Push to Godot Asset Library
+        uses: deep-entertainment/godot-asset-lib-action@v0.4.0
+        with:
+          username: ${{ secrets.GODOT_ASSET_LIBRARY_USERNAME }}
+          password: ${{ secrets.GODOT_ASSET_LIBRARY_PASSWORD }}
+          assetId: TODO
```
wjt, line 21: The underlying API does support creating a new asset rather than just updating an existing asset, but this GitHub Action (which I quickly read the source of – it doesn't appear to be about to steal our password) requires that it be specified. Fair enough.

dbnicholson: Presuming it gets approved, it should be [12617](https://godotengine.org/asset-library/asset/12617).

wjt:
Looking at the [API documentation](https://github.com/godotengine/godot-asset-library/blob/master/API.md#post-asset-post-assetid-post-asseteditid) I don't see any reason why this shouldn't work for an asset pending review.

dbnicholson (approved):
Seems like it should work!

.github/workflows/godot-asset-library.yaml
dbnicholson: As it turns out, that was just the ID for the submission I guess. The real ID is [3095](https://godotengine.org/asset-library/asset/3095). Fixup coming.

PR #22, opened 2024-06-13
Reduce first import errors

manuq:
https://phabricator.endlessm.com/T35471

### DIFF
--- a/addons/block_code/block_code_plugin.gd
+++ b/addons/block_code/block_code_plugin.gd
@@ -13,6 +13,10 @@ var old_feature_profile: String = ""
 
 
 func _enter_tree():
+	# Hack for `EditorInterface.open_scene_from_path()`, see: https://github.com/godotengine/godot/issues/86869
+	for _frame in range(10):
+		await get_tree().process_frame
+
 	main_panel = MainPanel.instantiate()
 	main_panel.undo_redo = get_undo_redo()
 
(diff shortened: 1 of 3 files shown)


wnbaum (changes requested):
Good that these reduce a lot of errors, but I don't really like how hacky they are, and having to preload a bunch of scripts instead of having the global names is kind of unfortunate.

I found that in the `.godot` folder, there is a file named `global_script_class_cache.cfg`. If this file doesn't exist when the editor is opened and the plugin is enabled, there will be errors when the global classes are being referenced for the first time. If someone installs and enables this plugin normally, these errors do not occur.

So, if there is a way to un-ignore this `.godot/global_script_class_cache.cfg` file for developers only, that would be a way to fix the errors (which come from a developer opening the project without a `.godot` folder and specifically the mentioned file not being generated yet).

manuq:
@wnbaum I think this is how other editor plugins are handling the current Godot bug. I borrowed from GDQuest Tour. See for reference: https://github.com/godotengine/godot/issues/81615

I think we should merge this while the upstream bug persists.

wnbaum:
Since this is only an issue for developers, and there's still a lot of errors happening that require a refresh I feel like we shouldn't merge yet.

Also, cloning the repository and adding the `global_script_class_cache.cfg` file wasn't actually enough to get rid of the errors. I had to copy the full `.godot` folder over. I think it's worth testing which files we need to include at some point.

But also if you really think it's worth it to merge then that's fine too :)

PR #25, opened 2024-06-13
Add EntryBlock class

wnbaum:
This felt like a good addition to have, since before the script generator was looking for a `block_name` for entry blocks, and generating the code itself, but now it asks the `EntryBlock` for the code, and appends it to the script that way.

`EntryBlock` inherits from `StatementBlock` (both scene and script), utilizing the statement block to generate the parameters UI programatically. This way we can expose things like signals by automatically creating an `EntryBlock` for them.

Originally we had been using `BasicBlock` as a template for entry blocks, but it is much nicer to have an `EntryBlock` that can actually generate a method signature.

### DIFF
--- a/addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
@@ -9,24 +9,14 @@ func generate_script_from_current_window(script_inherits: String = ""):
 
 	var blocks := current_window.get_children()
 
-	var ready_nodes: Array[BasicBlock] = []
-	var process_nodes: Array[BasicBlock] = []
-	var physics_process_nodes: Array[BasicBlock] = []
-	var signal_nodes: Array[StatementBlock] = []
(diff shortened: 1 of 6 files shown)


addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
```diff
@@ -9,65 +9,34 @@ func generate_script_from_current_window(script_inherits: String = ""):
...
 
 	var script: String = ""
 
 	script += "extends %s\n\n" % script_inherits
 
 	script += "var VAR_DICT := {}\n\n"
 
-	var node_groups = [["func _ready():", ready_nodes], ["func _process(_delta):", process_nodes], ["func _physics_process(_delta):", physics_process_nodes]]
+	for entry_block in entry_blocks:
+		script += entry_block.get_entry_statement() + "\n"
```
manuq, line 28: Yes!

manuq (approved):
Looks good!

PR #27, opened 2024-06-14
Add simple sound node

starnight:


starnight:
I do not know the path `addons/block_code/simple_nodes/simple_sound` is appropriate, or not.  Suggestion is welcomed.

Other questions:
* **stream** is an exposed property.  Should it be set in ready()?  Or, user can set it within Godot?
* Does the block programming support a parameter's default value?  For example the **Volume dB**.
* Does the block programming support multiple parameters?

Set this as a draft, due to the questions

wnbaum:
> I do not know the path `addons/block_code/simple_nodes/simple_sound` is appropriate, or not. Suggestion is welcomed.
> 
> Other questions:
> 
> * **stream** is an exposed property.  Should it be set in ready()?  Or, user can set it within Godot?
> * Does the block programming support a parameter's default value?  For example the **Volume dB**.
> * Does the block programming support multiple parameters?
> 
> Set this as a draft, due to the questions

- The path is correct, so that's good.
- I think eventually we could have a way for the user to set the stream/sound via blocks, but setting it from the inspector is fine for now, so you don't have to expose `stream`.
- Default parameters aren't implemented yet, but we are working on it.
- Yes, you can add multiple parameters to a block, for example this "Add NODE to GROUP" block:
```
b = BLOCKS["statement_block"].instantiate()
b.block_format = "Add {node: NODE} to group {group: STRING}"
b.statement = "{node}.add_to_group({group})"
signal_list.append(b)
```

addons/block_code/simple_nodes/simple_sound/simple_sound.gd
```diff
@@ -0,0 +1,33 @@
...
+
+static func get_custom_blocks() -> Array[BlockCategory]:
+	var b: Block
+
+	# Play the sound
+	var sound_list: Array[Block] = []
+	b = CategoryFactory.BLOCKS["statement_block"].instantiate()
+	b.block_type = Types.BlockType.EXECUTE
+	b.block_format = "Play the sound, Volume dB {db: INT}"
+	b.statement = "volume_db=db\n" + "play()"
```
wnbaum, line 28: Also, make sure the parameter to format is surrounded by curly braces `{}` or it won't be replaced by the value in the block.
```suggestion
	b.statement = "volume_db={db}\n" + "play()"
```

starnight: cool!  thanks!

manuq:
This will need an AudioStreamPlayer node with a BlockCode node attached to it with a single "Play the sound" block attached.

Maybe it's better to have the "Play sound" in every node and instantiate AudioStreamPlayer programatically. Something like:

```
var sound = AudioStreamPlayer.new()
add_child(sound)
var stream = load("res://../some.ogg")
sound.set_stream(stream)
# optionally parameters for volume and pitch:
# sound.volume_db = 1
# sound.pitch_scale = 1
sound.play()
```

starnight:
https://github.com/endlessm/block-programming-plugin/pull/36#discussion_r1642703128 gives me a good direction! :)

I think we can have 2 blocks:
* Load the audio file as the sound and add the sound as a child of the node: This can be used in `_ready()`.
* Find the child node sound, then play the sound with dB and pitch: This can be used in a signal handler, for example.

The sound can be passed by [add_child()](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-add-child) and [find_child()](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-find-child). The node's **name** is the key for finding.

Screenshot from 2024-06-18 17-51-13

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -235,6 +235,23 @@ static func get_general_categories() -> Array[BlockCategory]:
 	var input_list: Array[Block] = _get_input_blocks()
 	var input_category: BlockCategory = BlockCategory.new("Input", input_list, Color.SLATE_GRAY)
 
+	# Sound
+	var sound_list: Array[Block] = []
+
+	b = BLOCKS["parameter_block"].instantiate()
```
manuq, line 241:
```suggestion
	b = BLOCKS["statement_block"].instantiate()
```

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -235,6 +235,23 @@ static func get_general_categories() -> Array[BlockCategory]:
...
+	b.block_type = Types.BlockType.EXECUTE
+	b.block_format = "Load file {file_path: STRING} as sound {name: STRING}"
+	b.statement = "var sound = AudioStreamPlayer.new()\nsound.name = {name}\nsound.set_stream(load({file_path}))\nadd_child(sound)"
+	sound_list.append(b)
+
+	b = BLOCKS["parameter_block"].instantiate()
```
manuq, line 247:
```suggestion
	b = BLOCKS["statement_block"].instantiate()
```

manuq (changes requested):
In the UI these blocks should look with knobs at the top and bottom. So they need to instantiate "statement_block".

manuq:
> The sound can be passed by [add_child()](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-add-child) and [find_child()](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-find-child). The node's **name** is the key for finding.

This is looking great, but I wasn't able to make it work. Here is my attempt. I get a runtime error setting volume_db in a null instance.
Captura desde 2024-06-18 09-36-11

starnight:
Hmm!  I forget set the child node's owner: `sound.set_owner(self)`

Here is the test, and works now!
image

```gdscript
extends SimpleCharacter

var VAR_DICT := {}

func _ready():
	print('Hi Manuel!')
	add_to_group('Player')
	var signal_manager = get_tree().root.get_node_or_null("SignalManager")
	if signal_manager:
		signal_manager.broadcast_signal('Player', 'will_hi')
	var sound = AudioStreamPlayer.new()
	sound.name = 'testsound'
	sound.set_stream(load('res://test_game/score.ogg'))
	add_child(sound)
	sound.set_owner(self)

func _physics_process(delta):
	velocity = Input.get_vector("player_2_left", "player_2_right", "player_2_up", "player_2_down")*600
	move_and_slide()
	if Input.is_action_pressed("ui_right"):
		var sound = find_child('testsound')
		sound.volume_db = 0
		sound.pitch_scale = 1
		sound.play()


Debug script! (not saved)
```

addons/block_code/ui/picker/categories/category_factory.gd
starnight: sure!

addons/block_code/ui/picker/categories/category_factory.gd
starnight: sure!

manuq (approved):

PR #30, opened 2024-06-14
Block resources: Dehardcode scene paths

manuq:
Use the block class in the resources instead of the scene path. This decouples the block data from the block UI. Refactoring the scenes shouldn't break the block data.

### DIFF
--- a/addons/block_code/ui/block_canvas/block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/block_canvas.gd
@@ -7,7 +7,23 @@ const EXTEND_MARGIN: float = 800
 @onready var _window: Control = %Window
 @onready var _window_scroll: ScrollContainer = %WindowScroll
 
+var _block_scenes_by_class = {}
+
 signal reconnect_block(block: Block)
+
+
+func _ready():
+	_populate_block_scenes_by_class()
+
+
+func _populate_block_scenes_by_class():
+	for _class in ProjectSettings.get_global_class_list():
+		if not _class.base.ends_with("Block"):
+			continue
+		var _script = load(_class.path)
+		if not _script.has_method("get_scene_path"):
+			continue
+		_block_scenes_by_class[_class.class] = _script.get_scene_path()
 
 
 func add_block(block: Block) -> void:
@@ -34,14 +50,9 @@ func clear_canvas():
 		child.queue_free()
 
 
-#func load_canvas():
-#var save: SerializedBlockTreeNodeArray = ResourceLoader.load("user://test_canvas.tres")
-#for tree in save.array:
-#load_tree(_window, tree)
-
-
 func load_tree(parent: Node, node: SerializedBlockTreeNode):
-	var scene: Block = load(node.serialized_block.block_path).instantiate()
+	var _block_scene_path = _block_scenes_by_class[node.serialized_block.block_class]
+	var scene: Block = load(_block_scene_path).instantiate()
 	for prop_pair in node.serialized_block.serialized_props:
 		scene.set(prop_pair[0], prop_pair[1])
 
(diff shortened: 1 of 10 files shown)


addons/block_code/ui/blocks/control_block/control_block.gd
```diff
@@ -83,7 +83,11 @@ func get_serialized_props() -> Array:
 	return props
 
 
-func get_scene_path():
+func get_class():
+	return "ControlBlock"
```
wjt, line 87: Is there really no built-in way to ask a value for its class name?

Should these return `StringName` rather than `String`?

wjt: OK, I see that the built-in way is … `get_class()` which returns `String`, and ignores `class_name`.

https://docs.godotengine.org/en/4.2/classes/class_object.html#class-object-method-get-class

manuq: I will try with StringName and see how it goes!

As per the actual question, I couldn't find any for custom classes. A lot of discussion in issues like https://github.com/godotengine/godot/issues/21789

wjt: I think String is correct – you're overriding a method that returns `String`.

I have tried some experiments and I couldn't find a better way either. Sad.

manuq: @wjt it worked with StringName.

wjt: I think you should keep it as a String because https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-method-get-class has return type String. But maybe StringName is close enough to being a String…

wjt: Either that, or give it a different name. The advantage of defining a new method is that when we inevitably forget to implement it for some new block type, we'll get an error rather than falling back to the built in `Object.get_class`

manuq: @wjt yes good point, let's go with a new method to also avoid confusion.

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -7,9 +7,25 @@ const EXTEND_MARGIN: float = 800
...
 signal reconnect_block(block: Block)
 
 
+func _ready():
+	_populate_block_scenes_by_class()
+
+
+func _populate_block_scenes_by_class():
+	for _class in ProjectSettings.get_global_class_list():
+		if not _class.base.ends_with("Block"):
```
wjt, line 21: I don't like this being a substring match. Can we instead use `ClassDB.is_parent_class` https://docs.godotengine.org/en/stable/classes/class_classdb.html#class-classdb-method-is-parent-class to see if the class is a subclass of Block?

manuq: Good point! Checking...

manuq: I tried comparing like this:
```
		if not ClassDB.is_parent_class(_class.base, &"Block"):
			continue
```
And with "Block" (no &).. and it didn't work :(

manuq:
@wjt ready for review again, please note the last fixup.

addons/block_code/ui/blocks/control_block/control_block.gd
manuq: The other nice thing of using a separate method: we can mark it as static.

wnbaum:
@manuq I just went over everything and it looks good to me! Maybe you could do some kind of while loop to see if the parent class is `Block`, but what you have is good for now. Unfortunately `ClassDB` only works for built in classes. They should really have a more integrated system with `class_name`.

manuq:
> @manuq I just went over everything and it looks good to me! Maybe you could do some kind of while loop to see if the parent class is `Block`, but what you have is good for now. Unfortunately `ClassDB` only works for built in classes. They should really have a more integrated system with `class_name`.

Yeah. OK thanks, I'll squash the commits and merge.

PR #31, opened 2024-06-14
Input for all nodes

manuq:


manuq:
Note: this is on top of #30 . It allows things like:
Captura desde 2024-06-14 15-23-05

The list of blocks is overwelming, until we have dropdowns.

wnbaum (approved):
This is very cool! All three of the `pressed`, `just_pressed`, and `just_released` will be very useful I'm sure. We can definitely reduce the number of blocks once we have dropdowns.

manuq:
Actually this can be split from the "dehardcode" PR.

PR #32, opened 2024-06-14
Type casting

wnbaum:
This adds type casting to blocks. So if there is a defined way to cast from `INT` to `STRING`, you can snap an `INT` block to a `STRING` block, for example if you wanted to print out a math expression. The program will find the shortest way to convert an `INT` to a `STRING` and it will be added to the generated code.

I think this is nicer than having to do some sort of inheritance where we have a `NUMBER` class etc. Let me know what you think!


manuq (approved):
Looks good!

PR #33, opened 2024-06-14
Differentiate parameter input based on type

wnbaum:
The parameter input should change based on what is being inputted. For example, if we have a color input, we should show a color picker. For a Vector2 input we should have two inputs for each component ideally, though at the moment you have to input `0,0` for example.

I've implemented the color picker. We should also be able to set a default value for the parameter input. In `StatementBlock` where we create a parameter based on a string like `{test: STRING}`, we should be able to give it a default value like 
`{test: STRING = "Hello World!"}`.

We could also have a new type called `DROPDOWN` that exposes text options to the user to modify the block.


wnbaum:
Can now add default values to parameter inputs. Will be helpful for dropdown.

wnbaum:
Added dropdowns, condensing some of the logic blocks for example.

manuq:
> Added dropdowns, condensing some of the logic blocks for example.

This is looking great!

addons/block_code/types/types.gd
```diff
@@ -12,14 +12,16 @@ enum BlockType {
 	VECTOR2,
 	BOOL,
 	COLOR,
-	NODE
+	NODE,
+	OPTION
```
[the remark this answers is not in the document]
wnbaum: Sure, I was just reusing the parameter input generation we already had, because it ties into the code generation, and it basically is just a parameter input with some default text values. I suppose I can implement another component for this and add different syntax to create them in the `StatementBlock` formatting code if you think that's better.

addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
```diff
@@ -11,19 +11,72 @@ signal text_modified
...
+@onready var _x_line_edit := %XLineEdit
+@onready var _y_line_edit := %YLineEdit
+# Bool
+@onready var _bool_input := %BoolInput
+@onready var _bool_input_option := %BoolInputOption
+
+
+func set_raw_input(raw_input):
+	if option:
+		_panel.visible = false
```
manuq, line 41: You can save some indentation by returning earlier:
```
	if not option:
		return

	_panel.visible = false
	(...)

```

addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
```diff
@@ -11,19 +11,72 @@ signal text_modified
...
+			_x_line_edit.text = split[0]
+			_y_line_edit.text = split[1]
+		TYPE_BOOL:
+			_bool_input_option.select(raw_input)
+		_:
+			_line_edit.text = raw_input
+
+
+func get_raw_input():
+	if option:
```
manuq, line 65: Same here about the early return for indentation.

manuq (approved):
Looks great!

wnbaum:
Had to rebase again, should be good now.

addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
wnbaum: The diff looks kind of weird, but I can't return early because if `option` is false I want to do the normal `variant_type` logic.

PR #35, opened 2024-06-15
Add Pong demo

manuq:
And make it the main scene.
    
The paddles have 2 custom blocks for movement, borrowing from the SimpleCharacter example. But the speed parameter is a Vector2 so the movement can be limited to the Y axis.
    
The HUD has methods to change the scoring labels. For now a BlockCode node is added to the root node for setting arbitrary score as example.
    
All components have been copied from Moddable Pong and removing the logic. The ball properties were left for now.
    
The project window viewport is adjusted to match Moddable Pong. Also the gravity setting was set to zero.

https://phabricator.endlessm.com/T35504


starnight (approved):
LGTM!

PR #36, opened 2024-06-17
Add math nodes

starnight:
https://phabricator.endlessm.com/T35502

### DIFF
--- a/addons/block_code/simple_nodes/math/plus.gd
+++ b/addons/block_code/simple_nodes/math/plus.gd
@@ -1,0 +1,21 @@
+@tool
+class_name Plus
+extends Node
+
+func get_class():
+	return "Plus"
+
+static func get_custom_blocks() -> Array[BlockCategory]:
+	var b: Block
+
+	# Plus
+	var plus_list: Array[Block] = []
+	b = CategoryFactory.BLOCKS["statement_block"].instantiate()
+	b.block_type = Types.BlockType.EXECUTE
+	b.block_format = "{a: FLOAT} plus {b: FLOAT}"
+	b.statement = "return {a} + {b}"
+	plus_list.append(b)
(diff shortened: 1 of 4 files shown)


addons/block_code/simple_nodes/math/plus.gd
```diff
@@ -0,0 +1,21 @@
...
+	return "Plus"
+
+static func get_custom_blocks() -> Array[BlockCategory]:
+	var b: Block
+
+	# Plus
+	var plus_list: Array[Block] = []
+	b = CategoryFactory.BLOCKS["statement_block"].instantiate()
+	b.block_type = Types.BlockType.EXECUTE
+	b.block_format = "{a: FLOAT} plus {b: FLOAT}"
```
manuq, line 15: There are Math nodes already for integers at https://github.com/endlessm/block-programming-plugin/blob/38e74065955178be50d6c3a464f4c6f8614eef7d/addons/block_code/ui/picker/categories/category_factory.gd#L173

Ideally those should be overloaded to work for all possible Variants. Eg. it should be possible to multiply Vector2 by a float: `print(Vector2.ONE * 1.5)`.

manuq (changes requested):

addons/block_code/simple_nodes/math/plus.gd
starnight: ah!  I see!  Thanks!

starnight:
It already exists at [block-programming-plugin/addons/block_code/ui/picker/categories/category_factory.gd](https://github.com/endlessm/block-programming-plugin/blob/38e74065955178be50d6c3a464f4c6f8614eef7d/addons/block_code/ui/picker/categories/category_factory.gd#L173).  So, close this.

PR #37, opened 2024-06-17
instruction_tree: Remove unused generate_tree() method

wjt:
This method was introduced in 0b392c281a31d7f41031674debe7cc06e949fff5 but never used.


dbnicholson (approved):
I think that's right. We can always revert if that turns out to be needed.

PR #39, opened 2024-06-17
WIP: Strings as types

wnbaum:


manuq:
Honestly I would prefer to keep things as before in a custom ENUM rather than allowing arbitrary strings for the types.

Here is my alternative: https://github.com/endlessm/block-programming-plugin/pull/43

wjt:
Based on #43 having been merged, should we close this?

PR #40, opened 2024-06-17
Add tests

dbnicholson:
This adds the [GUT plugin](https://gut.readthedocs.io/en/latest/) to run unit tests and an example set of tests for `InstructionTree`. I've also included a GitHub Actions job to run the tests, but we'll see if it can be made to work correctly and reliably.

https://phabricator.endlessm.com/T35508


manuq (approved):
Looks good1

PR #41, opened 2024-06-17
Show Block Code editor when BlockCode node is selected

dylanmccall:
By default, when a node is selected, Godot shows the 2D or 3D editor corresponding to the selected node type. Showing the Block Code editor when a BlockCode node is selected will make the flow of the editor easier to understand.

### DIFF
--- a/addons/block_code/block_code_plugin.gd
+++ b/addons/block_code/block_code_plugin.gd
@@ -5,6 +5,8 @@ extends EditorPlugin
 const MainPanel := preload("res://addons/block_code/ui/main_panel.tscn")
 static var main_panel
 
+var editor_inspector: EditorInspector
+
 var old_feature_profile: String = ""
 
 const DISABLED_CLASSES := [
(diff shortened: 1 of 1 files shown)


addons/block_code/block_code_plugin.gd
```diff
@@ -91,6 +95,16 @@ func _exit_tree():
 			EditorInterface.set_current_feature_profile("")
 
 
+func _ready():
+	editor_inspector.connect("edited_object_changed", _on_editor_inspector_edited_object_changed)
+
+
+func _on_editor_inspector_edited_object_changed():
+	var block_code: BlockCode = editor_inspector.get_edited_object() as BlockCode
```
dbnicholson, line 103: I was not aware of Godot's behavior to set the casted value to `null` if [doesn't match the casted type](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/static_typing.html#type-casting). That's both neat and frightening.

dbnicholson (approved):
:clap: My godot knowledge is not high enough to know if this is the correct way to do it, but it certainly looks reasonable.

manuq:
This is great @dylanmccall !

PR #45, opened 2024-06-18
github: Publish test results

wjt:
This annotates the source location when a check fails, and adds a comment if at least one test has failed.

https://phabricator.endlessm.com/T35508

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -24,6 +24,9 @@ jobs:
   tests:
     name: Tests
     runs-on: ubuntu-latest
+    permissions:
+      checks: write
+      pull-requests: write
     steps:
     - name: Checkout
       uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/checks.yaml
```diff
@@ -37,4 +40,11 @@ jobs:
...
       run: |
-        godot --path . --headless --script addons/gut/gut_cmdln.gd -gexit
+        godot --path . --headless --script addons/gut/gut_cmdln.gd -gjunit_xml_file=test-report.xml -gexit
+    - name: Publish test results
+      uses: EnricoMi/publish-unit-test-result-action@v2
+      if: always()
+      with:
+        files: test-report.xml
+        comment_mode: failures
+        test_file_prefix: "-res://"
```
wjt, line 50: This doesn't seem to work; so the annotations don't get lined up with the source files.

wjt:
Not an improvement on just running the test.

PR #46, opened 2024-06-18
instruction_tree: Pass depth & output buffer as parameters

wjt:
I think this is clearer than using instance variables, which persist beyond the call to generate_text, to store state which is transient to any given call to generate_text. It's also shorter!


manuq (approved):
Nice!

PR #47, opened 2024-06-18
Add overview doc

wnbaum:
This adds a doc file that gives an overview of the flow of the program and explains the organization of the files.

Will add section for process of adding a new block.

https://phabricator.endlessm.com/T35448

### DIFF
--- a/docs/OVERVIEW.md
+++ b/docs/OVERVIEW.md
@@ -1,0 +1,70 @@
+# Block Programming Plugin
+
+## Data flow/Hierarchy
+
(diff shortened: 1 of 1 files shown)


manuq:
Looking good! Can we just remove the deprecated stuff?

wnbaum:
> Looking good! Can we just remove the deprecated stuff?

Yes! There was a point where it might have been useful (a sort of Picker for the "simple nodes"), but the design has since moved away from my custom scene editor idea.

dbnicholson (approved):
This is great! I think unless you have something you want to change right now, I'd just merge this now. It's very helpful as is and by definition it's going to need to be iterated on as we change the plugin.

wnbaum:
Sure, sounds good!

PR #49, opened 2024-06-19
Update project title

manuq:


project.godot
```diff
@@ -10,7 +10,7 @@ config_version=5
 
 [application]
 
-config/name="block-programming-plugin"
+config/name="Block coding (for learners)"
```
wjt, line 13: FWIW this file isn't part of what we will export to the asset library; it will only be seen by people developing the plugin itself.

dbnicholson: Indeed, but I like a nice title in the project list when hacking the plugin. Since it's only for developers, can we change it to `Block Coding` or `Block Coding Plugin` to better match the asset? I don't think the `(for learners)` part is particularly relevant to someone developing the plugin.

wjt: I found your argument convincing & refreshed the branch.

PR #54, opened 2024-06-19
Add ‘On body entered’ entry block for Area2D and RigidBody2D

wjt:
The block ‘compiles’ to a callback function, which needs to be connected
to the signal in question. To do this, add an additional code-generation
step that generates an _init() function to connect signals.

A more idiomatic approach would be to connect to the signals in the
_ready() function. This would involve merging additional code with the
user-supplied _ready() function, so is deferred for now. _init() works
fine.

https://phabricator.endlessm.com/T35510

### DIFF
--- a/addons/block_code/ui/blocks/parameter_block/parameter_block.gd
+++ b/addons/block_code/ui/blocks/parameter_block/parameter_block.gd
@@ -35,7 +35,7 @@ func _on_drag_drop_area_mouse_down():
 
 func get_serialized_props() -> Array:
 	var props := super()
-	props.append_array(serialize_props(["block_format", "statement"]))
+	props.append_array(serialize_props(["block_format", "statement", "variant_type"]))
 
 	var _param_input_strings: Dictionary = {}
 	for pair in param_name_input_pairs:
--- a/addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
+++ b/addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
@@ -55,7 +55,11 @@ func get_string() -> String:
 	var snapped_block: ParameterBlock = get_snapped_block() as ParameterBlock
 	if snapped_block:
 		var generated_string = snapped_block.get_parameter_string()
-		return Types.cast(generated_string, snapped_block.variant_type, variant_type)
+		if Types.can_cast(snapped_block.variant_type, variant_type):
+			return Types.cast(generated_string, snapped_block.variant_type, variant_type)
+		else:
+			push_warning("No cast from %s to %s; using '%s' verbatim" % [snapped_block, variant_type, generated_string])
+			return generated_string
 
 	var text: String = get_plain_text()
 
--- a/addons/block_code/ui/picker/categories/category_factory.gd
+++ b/addons/block_code/ui/picker/categories/category_factory.gd
@@ -348,7 +348,7 @@ static func get_built_in_categories(_class_name: String) -> Array[BlockCategory]
 	var class_cat: BlockCategory = category_from_property_list(prop_list, props, _class_name, Color.SLATE_GRAY)
 	block_list.append_array(class_cat.block_list)
 	class_cat.block_list = block_list
-	if props.size() > 0:
+	if block_list:
 		cats.append(class_cat)
 
 	return cats
--- a/addons/block_code/ui/picker/categories/category_factory.gd
+++ b/addons/block_code/ui/picker/categories/category_factory.gd
@@ -83,11 +83,6 @@ static func get_general_categories() -> Array[BlockCategory]:
 	b.statement = "print({text})"
 	test_list.append(b)
 
-	b = BLOCKS["entry_block"].instantiate()
-	b.block_format = "On body enter [body: NODE_PATH]"
-	b.statement = "func _on_body_enter(body):"
-	test_list.append(b)
-
 	var test_category: BlockCategory = BlockCategory.new("Test", test_list, Color("9989df"))
 
 	# Signal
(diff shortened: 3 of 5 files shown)


wjt:
In my working copy I deleted most of the blocks in the test game and recreated them to test this functionality; I've left that out here.

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -336,16 +334,31 @@ static func get_built_in_categories(_class_name: String) -> Array[BlockCategory]
...
 
 		"RigidBody2D":
-			# On body entered
+			var b = BLOCKS["entry_block"].instantiate()
+			b.block_format = "On body entered [body: NODE_PATH]"
+			# HACK: Blocks refer to nodes by path but the callback receives the node itself;
+			# convert to path
+			b.statement = "func _on_body_entered(_body: Node):\n\tvar body: NodePath = _body.get_path()"
+			b.signal_name = "body_entered"
+			block_list.append(b)
```
wjt, line 343: I actually don't think this will work for RigidBody2D. The [documentation for the `body_entered` signal says](https://docs.godotengine.org/en/stable/classes/class_rigidbody2d.html#class-rigidbody2d-signal-body-entered):

> Requires [contact_monitor](https://docs.godotengine.org/en/stable/classes/class_rigidbody2d.html#class-rigidbody2d-property-contact-monitor) to be set to true and [max_contacts_reported](https://docs.godotengine.org/en/stable/classes/class_rigidbody2d.html#class-rigidbody2d-property-max-contacts-reported) to be set high enough to detect all the collisions.

But those properties default to false and 0 respectively.

addons/block_code/ui/blocks/parameter_block/parameter_block.gd
```diff
@@ -35,7 +35,7 @@ func _on_drag_drop_area_mouse_down():
 
 func get_serialized_props() -> Array:
 	var props := super()
-	props.append_array(serialize_props(["block_format", "statement"]))
+	props.append_array(serialize_props(["block_format", "statement", "variant_type"]))
```
manuq, line 38: Oh good catch!

addons/block_code/ui/picker/categories/category_factory.gd
wjt: (Of course the game developer can set those properties but it's not obvious from the block that you have to do so.)

addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
```diff
@@ -55,7 +55,11 @@ func get_string() -> String:
 	var snapped_block: ParameterBlock = get_snapped_block() as ParameterBlock
 	if snapped_block:
 		var generated_string = snapped_block.get_parameter_string()
-		return Types.cast(generated_string, snapped_block.variant_type, variant_type)
+		if Types.can_cast(snapped_block.variant_type, variant_type):
+			return Types.cast(generated_string, snapped_block.variant_type, variant_type)
+		else:
+			push_warning("No cast from %s to %s; using '%s' verbatim" % [snapped_block, variant_type, generated_string])
+			return generated_string
```
manuq, line 62: Yes a warning and continue verbatim is much better.

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -336,16 +334,31 @@ static func get_built_in_categories(_class_name: String) -> Array[BlockCategory]
 			props = ["modulate"]
 
 		"RigidBody2D":
-			# On body entered
+			var b = BLOCKS["entry_block"].instantiate()
+			b.block_format = "On body entered [body: NODE_PATH]"
```
manuq, line 338: What about changing the sentence to:
```suggestion
			b.block_format = "On body [body: NODE_PATH] entered"
```

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -336,16 +334,31 @@ static func get_built_in_categories(_class_name: String) -> Array[BlockCategory]
...
+			# HACK: Blocks refer to nodes by path but the callback receives the node itself;
+			# convert to path
+			b.statement = "func _on_body_entered(_body: Node):\n\tvar body: NodePath = _body.get_path()"
+			b.signal_name = "body_entered"
```
manuq, line 342: It would be nice to have an "On body [body] exited" block too by connecting to the body_exited signal.

manuq (approved):

addons/block_code/ui/picker/categories/category_factory.gd
manuq: Yes, at least we need to make sure the contact monitor is enabled in the demo Pong project.

addons/block_code/ui/picker/categories/category_factory.gd
wjt: Or even "On [body: NODE_PATH] entered"?

addons/block_code/ui/picker/categories/category_factory.gd
wjt: Pushed an amended patch that defines that.

addons/block_code/ui/picker/categories/category_factory.gd
wjt: Pushed an amended patch.

addons/block_code/ui/blocks/parameter_block/parameter_block.gd
wnbaum: Found this in my ParameterInput PR too, hah.

wnbaum (approved):
Looks good to me! Working as expected on my end. Great work!!

PR #59, opened 2024-06-19
Fix base class identification (for Picker)

wnbaum:
This is just a really small fix for an error that was happening in the pong game. For previous classes with custom blocks, like `SimpleCharacter`, you had to define a method that specified the base class so the picker would know which blocks to populate itself with.

However, there is a function `script.get_instance_base_type()` that gets this automatically from whichever class the script extends. Much nicer than having to define a function for this every time.

### DIFF
--- a/addons/block_code/simple_nodes/simple_character/simple_character.gd
+++ b/addons/block_code/simple_nodes/simple_character/simple_character.gd
@@ -13,10 +13,6 @@ func get_custom_class():
 	return "SimpleCharacter"
 
 
-static func get_base_class():
-	return "CharacterBody2D"
-
-
 static func get_exposed_properties() -> Array[String]:
 	return ["position"]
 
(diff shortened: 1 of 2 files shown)


manuq:
Ah yes, I noticed this in the paddle blocks too. Paddles inherit from CharacterBody2D but they weren't getting Modulate from the parent CanvasItem class, position from the Node2D class, etc. Please add the explanation in the commit message, not in the PR message, then merge.

wjt:
I have reworded the commit message with your words, @wnbaum.

If you open a pull request for a single commit, its commit message will be used as the PR's title & body. So there is no extra overhead to writing a longer commit message – and there is an upside that when someone is looking through the commit history later to understand a change, the full explanation is right there.

wnbaum:
> I have reworded the commit message with your words, @wnbaum.
> 
> If you open a pull request for a single commit, its commit message will be used as the PR's title & body. So there is no extra overhead to writing a longer commit message – and there is an upside that when someone is looking through the commit history later to understand a change, the full explanation is right there.

Great! Good to know, thank you!

PR #60, opened 2024-06-19
Basic functional pong game

wnbaum:
I thought it would be fun to take a stab at implementing a very basic version of the pong game to see what other blocks/improvements we would need to implement.

I added scoring, and resetting the ball on score. For resetting the ball, interestingly enough, setting the `position` of a `RigidBody` does not work like a normal `Node2D`, you have to query the physics server. So I added a block for "Set Physics Position", which honestly isn't very clear for a beginner.

One thing I did find in making this is that we should think about making the variables more streamlined. Typing the variable name out every time is annoying without being able to copy blocks, and we should have a block for "Change variable by", so I don't have to use the "get variable" block every time.

I also found a small bug in `StatementBlock` where `TYPE_NIL` is technically equivalent to `false`, so it was being considered null.


addons/block_code/ui/blocks/statement_block/statement_block.gd
```diff
@@ -116,7 +116,7 @@ static func format_string(parent_block: Block, attach_to: Node, string: String,
 		var param_input: ParameterInput = preload("res://addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.tscn").instantiate()
 		param_input.name = "ParameterInput%d" % start  # Unique path
 		param_input.placeholder = param_name
-		if param_type:
+		if param_type != null:
```
manuq, line 119: :+1:

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -362,6 +362,11 @@ static func get_built_in_categories(_class_name: String) -> Array[BlockCategory]
 				b.signal_name = "body_%s" % [verb]
 				block_list.append(b)
 
+			var b = BLOCKS["statement_block"].instantiate()
+			b.block_format = "Set Physics Position {position: VECTOR2}"
+			b.statement = "PhysicsServer2D.body_set_state(get_rid(),PhysicsServer2D.BODY_STATE_TRANSFORM,Transform2D.IDENTITY.translated({position}))"
```
manuq, line 367: I wonder why this is not needed in Moddable Pong.

I think because what we do for respawning the ball is (from memory):
- Duplicate the node
- Queue free the old one
- Set a random position to the new one, and finally call add_child

addons/block_code/ui/picker/picker.gd
```diff
@@ -25,7 +25,7 @@ func bsd_selected(bsd: BlockScriptData):
 
 	var parent_class: String
 	if found_simple_class_script:
-		parent_class = found_simple_class_script.get_base_class()
+		parent_class = str(found_simple_class_script.get_instance_base_type())
```
manuq, line 28: Can you explain this change? What is this doing and why is it needed? Ideally in each commit message.

pong_game/goal_area.tscn
```diff
@@ -1,9 +1,10 @@
-[gd_scene load_steps=2 format=3 uid="uid://vc7u0ljq02eh"]
+[gd_scene load_steps=2 format=3 uid="uid://fhoapg3anjsu"]
 
 [sub_resource type="RectangleShape2D" id="RectangleShape2D_8q7gx"]
 size = Vector2(128, 1080)
 
 [node name="GoalArea" type="Area2D"]
+collision_mask = 3
```
manuq, line 7: :+1:

manuq (changes requested):
You did Pong! So cool! I left a comment about the implementation but we can iterate after merging. Feel free to merge, my only request is to explain the changes in the commit message body.

addons/block_code/ui/picker/categories/category_factory.gd
wnbaum: Yes, I think setting the position this way is fine too, since we don't have node instantiation yet. This seemed like the simplest way.

Also, we should add random blocks in the future! I remember using them very often in Scratch, and they introduced me to the world of procedural generation :)

wnbaum:
Rebased to incorporate sound nodes in the pong scene.

Should we move it into an `examples` folder in the addon?

addons/block_code/ui/picker/picker.gd
wnbaum: This was explained in #59.

dbnicholson (changes requested):
I like it being in an examples directory in the plugin so that someone who installs the plugin can play with a fully formed game and see how the plugin is intended to be used.

Could you also add something to the commit message like @manuq requested? You can probably just copy in the text you put in the PR.

wnbaum:
Sure, will do!

dbnicholson (approved):
Works for me. I was thinking maybe we should leave test_game as a complete playground, but I don't think there's a ton of benefit.

PR #62, opened 2024-06-20
Add a "Send signal to node" block

dylanmccall:
In #60, I wasn't a fan of us needing to reset _all_ of the ball nodes when a player scores, so I'm proposing a "send signal to node" block. This requires a little bit of tweaking so we aren't defining `var signal_manager` in either of the send signal blocks, which makes it possible to include multiple of them in the same context.


wnbaum:
This is really good! I also thought having this block would be useful when I was making the pong game.

Signals aren't currently working pending [a fix](https://github.com/endlessm/godot-block-coding/pull/60/commits/45271066031fd3f14922f1af0b54bc19ee1c1966) in the pong game PR which I hopefully will get merged shortly.

wnbaum (approved):
Still working great! Thanks!

PR #63, opened 2024-06-20
Add getting started section to README

wnbaum:
This section of documentation explains how to set up and begin using the plugin.

We should add the AssetLib link once it exists!


dbnicholson (approved):
:+1:

PR #64, opened 2024-06-20
Categories rework

dbnicholson:
This reworks the block category handling to (hopefully) make it more flexible and intuitive. The primary changes are:

* Store the desired category as a property of the block and then generate the categories from one big list of blocks. This is very much internal, but it allows defining blocks pretty much arbitrarily while maintaining the desired grouping. Previously you added categories and there was a process to merge duplicate categories. The end result is the same but I think this is a bit clearer and more efficient. This is [block: Make category a property of Block](https://github.com/endlessm/godot-block-coding/commit/115c92b46b2883f8c65ff5340feed32713de4dfa). It also adds an `order` property for categories so that you can take the arbitrary block list, group them into categories, and then sort the categories so they're displayed in the desired order. Previously the order was handled by the order the category was added.

* Change the categories of blocks created from classes so they go into action categories rather than class based categories. Feel free to bikeshed the categories, colors and assignments I made.

### DIFF
--- a/addons/block_code/examples/pong_game/paddle.gd
+++ b/addons/block_code/examples/pong_game/paddle.gd
@@ -11,23 +11,23 @@ static func get_exposed_properties() -> Array[String]:
 	return ["position"]
 
 
-static func get_custom_blocks() -> Array[BlockCategory]:
+static func get_custom_blocks() -> Array[Block]:
 	var b: Block
+	var block_list: Array[Block] = []
 
 	# Movement
-	var movement_list: Array[Block] = []
 	b = CategoryFactory.BLOCKS["statement_block"].instantiate()
 	b.block_type = Types.BlockType.EXECUTE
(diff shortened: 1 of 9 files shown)


dbnicholson:
Since I started the branch, pong and the overview docs were added. I've updated those now.

dbnicholson:
I'm gonna have to rebase again since I merged #60.

dbnicholson:
Ready again.

wnbaum (approved):
Looks good to me! Fantastic work!

I think there are a few other things we can do for categories, but should be done later:
1. Have support for subcategories (e.g. Movement/PhysicsBody2D would have "Set Physics Position" block). But they would stay in the same place with the same color.
2. Redo colors and choices for category names (according to Manuel's spec), and modify the pong game block scripts which still have the old blocks.

PR #65, opened 2024-06-21
README: improve structure

cassidyjames:
- Correct main heading
- Correct (and link to) EOSF
- Create a couple of more clear sections
- Consistently **bold** UI terms
- Consistent heading capitalization
- Link to tools that inspired us
- Add a sentence about learners familiarizing themselves with the Godot UI through our plugin
- Use the term "select" instead of "click" (minor nit, but is more inclusive of touch plus keyboard navigation—and more consistent with our OS docs)


starnight (approved):
LGTM!

PR #66, opened 2024-06-21
Updated logo

dbnicholson:
Let's make this WIP logo the project icon and iterate on the main branch from there. Note that the asset library icon refers to https://raw.githubusercontent.com/endlessm/godot-block-coding/logo/plugin-logo.png, so that will have to be updated when this is merged. In fact, I probably broke it with the rename in the last commit.


dbnicholson:
Indeed, I've broken the listing. I'm going to close this PR and use a new branch and change this branch so it keeps supporting the current asset library URL.

PR #70, opened 2024-06-21
When generating a script, merge duplicate entry blocks

dylanmccall:
If the user places multiple of the same entry block, all of the entry blocks should run. To support this, we will group entry blocks by entry statement.


wnbaum:
This is actually how I had it a while ago, but you did it a lot cleaner!

Code looks good to me and working in editor as expected. Great stuff!

PR #72, opened 2024-06-21
Clean up redundant code with multiple entry blocks

dylanmccall:
Forgot to commit this in #70 :)


wnbaum (approved):
LGTM! No more unnecessary pass or signal connect statements being generated. Nice!

PR #75, opened 2024-06-24
Reset snap point variables in DragManager.drag_end

dylanmccall:
This addresses a bunch of little things related to drag and drop in the editor. Importantly, I fixed two small bugs that occur when snapped nodes are separated from each other. There's also a bit of code cleanup, originally inspired by `previewing_snap_point` and `preview_block` in `DragManager` not being in sync with each other. In the end, I put those concepts into a separate `DragManager.Drag` class which manages a single drag & drop operation. That also provided a way to better organize our code related to filtering and sorting snap points, now using the stuff built in to `Array`.

I also ended up adding a change related to https://phabricator.endlessm.com/T35525, making a block fade out slightly if it is going to be deleted.

-----

https://phabricator.endlessm.com/T35524

### DIFF
--- a/addons/block_code/drag_manager/drag_manager.gd
+++ b/addons/block_code/drag_manager/drag_manager.gd
@@ -42,7 +42,6 @@ class Drag:
 		assert(block.get_parent() == null)
 
 		add_child(block)
-		block.on_canvas = false
 		block.position = -offset
 
 		_block = block
@@ -92,7 +91,7 @@ class Drag:
 			push_error("Warning: snap point %s does not reference its parent block." % _snap_point)
 			return false
 
-		if not _snap_point.block.on_canvas:
+		if not _block_canvas.is_ancestor_of(_snap_point):
 			# We only snap to blocks on the canvas:
 			return false
 
--- a/addons/block_code/drag_manager/drag_manager.gd
+++ b/addons/block_code/drag_manager/drag_manager.gd
@@ -59,7 +59,7 @@ class Drag:
 
 	func _remove_block():
 		target_snap_point = null
-		_block.g()
+		_block.queue_free()
 
 	func _place_block():
 		var canvas_rect: Rect2 = _block_canvas.get_global_rect()
--- a/addons/block_code/drag_manager/drag_manager.gd
+++ b/addons/block_code/drag_manager/drag_manager.gd
@@ -22,7 +22,9 @@ class Drag:
 		get:
 			return action
 		set(value):
-			action = value
+			if action != value:
+				action = value
+				_update_action_hint()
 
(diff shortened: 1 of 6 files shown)


manuq:
Is this one a separate snap issue? I can see it in main and in this PR:

[Grabación de pantalla desde 2024-06-24 16-15-07.webm](https://github.com/endlessm/godot-block-coding/assets/83944/82af0521-6f0a-4c74-8d8f-b3b9c36ea670)

Basically:
- Separate blocks from the entry block
- Move the entry block
- Try to drag the separated blocks again

The first 2 separated blocks can't be dragged.

dylanmccall:
> Is this one a separate snap issue? I can see it in main and in this PR:
> 
>  Grabacion.de.pantalla.desde.2024-06-24.16-15-07.webm 
> 
> Basically:
> 
> * Separate blocks from the entry block
> * Move the entry block
> * Try to drag the separated blocks again
> 
> The first 2 separated blocks can't be dragged.

It's an unrelated issue. If I start dragging a block but let it snap back to the original snap point, then start dragging it a _second_ time, it won't snap to that snap point again. (Which is a problem because the blocks still end up looking like they're snapped together even though they aren't). I think that issue you spotted is unrelated, but I'm continuing to fiddle with this code in a separate cleanup right now so I'll see if I can solve that at the same time.

dylanmccall:
Oh, I figured that one out, and it does sort of relate to my other cleanups here. It's related to the block still reserving all of that space it used to occupy so clicks weren't falling through to the nodes which were detached and pushed down in the z stack.

I'll stuff this all in the same PR so we don't have to merge things in order.

addons/block_code/drag_manager/drag_manager.gd
```diff
@@ -9,111 +9,197 @@ signal block_modified
...
+			return _block
+		elif action == DragAction.REMOVE:
+			_remove_block()
+			return null
+		else:
+			return null
+
+	func _remove_block():
+		target_snap_point = null
+		_block.g()
```
starnight, line 62: When the block becomes removed, it shows error:
`res://addons/block_code/drag_manager/drag_manager.gd:62 - Invalid call. Nonexistent function 'g' in base 'MarginContainer (EntryBlock)'.`

starnight (changes requested):

addons/block_code/drag_manager/drag_manager.gd
dylanmccall: Oh that's so weird! What did I do to that commit? :/ Thanks for spotting. I fixed that in the newest commit :)

manuq:
Thank you both! Dragging works much better now!

PR #76, opened 2024-06-24
Add camera to fit default project size

manuq:
The Godot project size defaults to 1152x648 which is 60% of the viewport in Pong. So adding a camera with 0.6 zoom makes the game fit in the default viewport. And makes it look odd in other sizes.

Also now the HUD layer follows the viewport size.

https://phabricator.endlessm.com/T35515


wnbaum:
Tested on new project and looks good to me!

PR #77, opened 2024-06-24
block_category_display: Make categories buttons to show/hide blocks

dbnicholson:
When all the categories are shown, you may need to do quite a bit of scrolling to find the block you want or just to discover what's there. This changes the category display so that each category is a button that can show or hide its blocks. All blocks are initially hidden to be able to see all the available categories.

The expander is implemented as a flat button with a triangle icon. The icon comes from an atlas with two states corresponded to expanded or unexpanded. Godot does provide the Tree class with similar functionality, but it's a bit complex to use. More importantly TreeItems are their own class that don't derive from Node. That would break the current Block usage.

The SVG with the triangle drawings is included should any editing be needed, but it's not used in the scene.

### DIFF
--- a/addons/block_code/ui/picker/categories/block_category_display.gd
+++ b/addons/block_code/ui/picker/categories/block_category_display.gd
@@ -2,22 +2,53 @@
 class_name BlockCategoryDisplay
 extends MarginContainer
 
+signal category_expanded(value: bool)
+
 var category: BlockCategory
 
-@onready var _label := %Label
+@onready var _button := %Button
+@onready var _blocks_container := %BlocksContainer
 @onready var _blocks := %Blocks
 @onready var _background := %Background
 
+@onready var _icon_collapsed := EditorInterface.get_editor_theme().get_icon("GuiTreeArrowRight", "EditorIcons")
+@onready var _icon_expanded := EditorInterface.get_editor_theme().get_icon("GuiTreeArrowDown", "EditorIcons")
 
-func _ready():
-	if category:
-		_label.text = category.name
-		_background.color = category.color.darkened(0.7)
+var expanded: bool:
+	set = _set_expanded
+
+
+func _set_expanded(value: bool):
+	expanded = value
+
+	_blocks_container.visible = expanded
+	if expanded:
+		_button.icon = _icon_expanded
+		_background.color = category.color.darkened(0.5)
+		_background.color.a = 0.3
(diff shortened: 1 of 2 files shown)


dbnicholson:
Here's a quick demo.

[Screencast from 2024-06-24 12-09-46.webm](https://github.com/endlessm/godot-block-coding/assets/6495518/93629fe0-ec7f-4280-b812-ea0d1c79bcf2)

manuq (approved):
Looks good and I think the change is very welcome!

One small thing: I wonder if editor icons from the default theme can be used. I just spotted [this icon previewer addon](https://godotengine.org/asset-library/asset/374) and does this to fetch editor icons: https://github.com/Xrayez/godot-editor-icons-previewer/blob/gd3/addons/editor_icon_previewer/editor_icon_previewer.gd#L51-L59

dbnicholson:
> One small thing: I wonder if editor icons from the default theme can be used. I just spotted [this icon previewer addon](https://godotengine.org/asset-library/asset/374) and does this to fetch editor icons: https://github.com/Xrayez/godot-editor-icons-previewer/blob/gd3/addons/editor_icon_previewer/editor_icon_previewer.gd#L51-L59

Oh, that's a good idea. I'll see if something is in there. Certainly drawing icons is not something I should normally be doing.

dbnicholson:
> > One small thing: I wonder if editor icons from the default theme can be used. I just spotted [this icon previewer addon](https://godotengine.org/asset-library/asset/374) and does this to fetch editor icons: https://github.com/Xrayez/godot-editor-icons-previewer/blob/gd3/addons/editor_icon_previewer/editor_icon_previewer.gd#L51-L59
> 
> Oh, that's a good idea. I'll see if something is in there. Certainly drawing icons is not something I should normally be doing.

I had to find a version in a branch that works with 4.2, but then I was able to find `GuiTreeArrowRight` and `GuiTreeArrowDown`. Presumably those are used in `Tree`. I also changed to try out @dylanmccall's suggestion that the color be more vibrant when in collapsed state.

dbnicholson:
New demo:
[Screencast from 2024-06-24 16-35-36.webm](https://github.com/endlessm/godot-block-coding/assets/6495518/c68f71e8-0249-4571-8468-f922dfc250d5)

starnight (approved):
I like this one!

PR #79, opened 2024-06-24
Add scope to blocks

wnbaum:
Made it so that parameter blocks copy-dragged from an `EntryBlock` can only be placed below entry blocks with the same method signature.

Also darken the block trees that cannot be snapped to by the current dragged block.

https://phabricator.endlessm.com/T35521

### DIFF
--- a/addons/block_code/drag_manager/drag_manager.gd
+++ b/addons/block_code/drag_manager/drag_manager.gd
@@ -108,12 +108,18 @@ class Drag:
 
 		# Check if any parent node is this node
 		var parent = _snap_point
+		var top_block
 		while parent is SnapPoint:
 			if parent.block == _block:
 				return false
 
+			top_block = parent.block
 			parent = parent.block.get_parent()
-
+			
+		if _block.scope != "":
+			if top_block is EntryBlock and _block.scope != top_block.get_entry_statement():
+				return false
+				
 		return true
 
 	func sort_snap_points_by_distance(a: SnapPoint, b: SnapPoint):
@@ -161,7 +167,6 @@ func _ready():
 
 func _process(_delta):
 	_update_drag_position()
-
 
 func _update_drag_position():
 	if not drag:
@@ -237,7 +242,13 @@ func connect_block_canvas_signals(block: Block):
 		var statement_block := block as StatementBlock
 		for pair in statement_block.param_name_input_pairs:
 			var param_input: ParameterInput = pair[1]
-			var b := param_input.get_snapped_block()
-			if b:
-				if b.drag_started.get_connections().size() == 0:
-					b.drag_started.connect(copy_picked_block_and_drag)
+			var copy_block := param_input.get_snapped_block()
+			if copy_block:
+				if copy_block.drag_started.get_connections().size() == 0:
+					copy_block.drag_started.connect(func(b: Block): drag_copy_parameter(b, block))
+
+
+func drag_copy_parameter(block: Block, parent: Block):
+	if parent is EntryBlock:
+		block.scope = parent.get_entry_statement()
+	copy_picked_block_and_drag(block)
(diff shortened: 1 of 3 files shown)


manuq (approved):
Looks good! Can you transfer the information in the PR description to each commit message body?

wnbaum:
> Looks good! Can you transfer the information in the PR description to each commit message body?

Yes! I'll be better at doing that in the future ;)

wnbaum:
Ok, should be better now @manuq 

I just realized however that we shouldn't just check the scope of the dragged block, but also the scope of the dragged block's child blocks. That way you couldn't drag a child block out of it's scope. Should I add that to this PR?

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -115,3 +115,20 @@ func find_snaps(node: Node) -> Array:
...
 	return snaps
+
+
+func set_scope(scope: String):
+	for block in _window.get_children():
+		if block is EntryBlock:
+			var entry_block = block as EntryBlock
+
+			if scope != entry_block.get_entry_statement():
+				entry_block.modulate = Color(0.5, 0.5, 0.5, 1)
```
manuq, line 126: As a general advice, at least in languages that have semantic indentation (Python, GDScript) it is better to return early to avoid nesting and thus very indented code. In a for loop you can do that with break/continue. In a function you can do that with return. Example:
```
func set_scope(scope: String):
	for block in _window.get_children():
		if block is not EntryBlock:
			continue
		var entry_block = block as EntryBlock
		(...)
```

manuq (changes requested):

manuq:
> Ok, should be better now @manuq
> 
> I just realized however that we shouldn't just check the scope of the dragged block, but also the scope of the dragged block's child blocks. That way you couldn't drag a child block out of it's scope. Should I add that to this PR?

I think it's fine to merge now after fixing conflicts and my request above. But if you think it's better to do the children blocks scoping now, that's fine with me too.

wnbaum:
Alright, I actually decided to just implement checking the child block scope as well. There should be no case where block trees will have blocks from different scopes.

Definitely gained some solid merging experience from rebasing with all of Dylan's new dragging/snapping code 💪

addons/block_code/drag_manager/drag_manager.gd
```diff
@@ -237,7 +257,26 @@ func connect_block_canvas_signals(block: Block):
...
 		for pair in statement_block.param_name_input_pairs:
 			var param_input: ParameterInput = pair[1]
-			var b := param_input.get_snapped_block()
-			if b:
-				if b.drag_started.get_connections().size() == 0:
-					b.drag_started.connect(copy_picked_block_and_drag)
+			var copy_block := param_input.get_snapped_block()
+			if copy_block:
+				if copy_block.drag_started.get_connections().size() == 0:
+					copy_block.drag_started.connect(func(b: Block): drag_copy_parameter(b, block))
```
manuq, line 263:
```suggestion
			if copy_block == null:
				continue
			if copy_block.drag_started.get_connections().size() == 0:
				copy_block.drag_started.connect(func(b: Block): drag_copy_parameter(b, block))
```

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -121,3 +121,24 @@ func find_snaps(node: Node) -> Array:
...
+
+		if block is EntryBlock:
+			if scope == block.get_entry_statement():
+				valid = true
+		else:
+			var tree_scope := DragManager.get_tree_scope(block)
+			if tree_scope == "" or scope == tree_scope:
+				valid = true
+
+		if !valid:
```
manuq, line 138: The `!` negation is deprecated in GDScript. Use `not`:
```suggestion
		if not valid:
```

manuq (changes requested):
Almost there!

manuq (approved):
@wnbaum looks good now! Feel free to squash the last commit and merge.

PR #85, opened 2024-06-26
github: Drop the publish job's action type check

starnight:
Current publish job is only triggered by published release. But, it is guarded by "released" action. The mismatched condition blocks the jobs. Drop the publish job's action type check to avoid this issue.


wjt:
This to me looks necessary but not sufficient. Rereading the documentation at https://github.com/deep-entertainment/godot-asset-lib-action?tab=readme-ov-file#example-usage I don't know why I included this check. But I see that we need to provide a template for the submission as well. I'll look at adding that.

PR #86, opened 2024-06-26
Update index number for categories order

manuq:
So they start at 10 and are separated by 10. Also 2 categories were sharing the same
number by mistake.

### DIFF
--- a/addons/block_code/ui/picker/categories/category_factory.gd
+++ b/addons/block_code/ui/picker/categories/category_factory.gd
@@ -16,102 +16,102 @@ const BUILTIN_PROPS: Dictionary = {
 	"Lifecycle":
 	{
 		"color": Color("ec3b59"),
-		"order": 000,
+		"order": 10,
 	},
 	"Transform | Position":
 	{
 		"color": Color("4b6584"),
-		"order": 010,
+		"order": 20,
 	},
 	"Transform | Rotation":
 	{
 		"color": Color("4b6584"),
-		"order": 011,
+		"order": 30,
 	},
 	"Transform | Scale":
 	{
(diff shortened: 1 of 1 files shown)


dbnicholson (approved):
This is fine, but I actually think we should simplify this a bit. I did these orderings to maintain exactly the list that was there before. The categories are sorted by order and then alphabetically if they have the same order. Therefore, I think we should just let the alphabetical part take more precedence and only lightly use the order value. I.e., make all of these `5` all the other space more or less reserved for custom categories to go to the top or bottom of the list. Or maybe we just drop the order and always sort alphabetically.

manuq:
> This is fine, but I actually think we should simplify this a bit. I did these orderings to maintain exactly the list that was there before. The categories are sorted by order and then alphabetically if they have the same order. Therefore, I think we should just let the alphabetical part take more precedence and only lightly use the order value. I.e., make all of these `5` all the other space more or less reserved for custom categories to go to the top or bottom of the list. Or maybe we just drop the order and always sort alphabetically.

For sorting alphabetically you mean in code or user facing? User facing I think is a good idea to have the more relevant categories at the top, and less relevant at the bottom. Just like Scratch or Microsoft Arcade do. See https://en.scratch-wiki.info/wiki/Block_Categories

dbnicholson:
> For sorting alphabetically you mean in code or user facing? User facing I think is a good idea to have the more relevant categories at the top, and less relevant at the bottom. Just like Scratch or Microsoft Arcade do. See https://en.scratch-wiki.info/wiki/Block_Categories

Probably right. I think we might be able to add some constants and helpers so that the magic order values don't need to be maintained like this.

PR #88, opened 2024-06-26
Populate more fields of plugin.cfg

wjt:
These are shown in the plugins section of the project settings.

The author and description fields can be populated statically, but the version number cannot unless we manually update it for each release. Instead, lean on a feature of `git archive` where magic placeholders in specified files can be substituted for strings based on the current commit, crucially including the output of `git describe` which will return a tag name if the current commit is an annotated tag.

v0.2.0 is annotated while v0.1.0 was not.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,5 @@
 /addons               !export-ignore
 /addons/block_code    !export-ignore
 /addons/block_code/** !export-ignore
+
+/addons/block_code/plugin.cfg export-subst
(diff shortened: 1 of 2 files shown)


wjt:
If you check out the project from Git you will see the following ugly version string in the settings:

image

Here is the result when I import a [snapshot archive of this branch](https://github.com/endlessm/godot-block-coding/archive/refs/heads/T35503-plugin.cfg-fields.zip) into another project

image

`v0.2.0-5-gc885832` decodes to:

- Closest tag is `v0.2.0` (as an aside: I dislike the `v` prefix)
- `5` indicates that it is 5 commits ahead of that tag
- `gc885832` indicates that the commit is c885832

manuq (approved):

PR #89, opened 2024-06-26
Add simple character to node dialog

wnbaum:
The SimpleCharacter should be available in the node creation dialog box!

### DIFF
--- a/addons/block_code/block_code_plugin.gd
+++ b/addons/block_code/block_code_plugin.gd
@@ -14,7 +14,6 @@ const DISABLED_CLASSES := [
 	"DragManager",
 	"InstructionTree",
 	"EditorInterfaceAccess",
-	"SimpleCharacter",
 	"Types",
 	"BasicBlock",
 	"Block",


manuq (changes requested):
Instantiating it as a scene worked fine for me yesterday. But if you want to add it as a node, is OK with me. Just make sure to do the same to SimpleScoring for consistency.

Oh and remainder to add the details in the commit message, not in the PR. We do that so the info remains in the git history, not just in github.

wnbaum:
> Instantiating it as a scene worked fine for me yesterday. But if you want to add it as a node, is OK with me. Just make sure to do the same to SimpleScoring for consistency.

The SimpleScoring is already available in the picker, and I'd just rather the learner not have to go into the source code of the plugin if they don't have to.

> Oh and remainder to add the details in the commit message, not in the PR. We do that so the info remains in the git history, not just in github.

And yes! I'll do that. I just thought it was pretty self explanatory :)

manuq:
Oh so this is removing simple character from a list of filtered nodes. Merging then.

PR #90, opened 2024-06-27
category_factory: Add tooltips into Lifecycle and Sounds blocks

starnight:
https://phabricator.endlessm.com/T35529

### DIFF
--- a/addons/block_code/ui/picker/categories/category_factory.gd
+++ b/addons/block_code/ui/picker/categories/category_factory.gd
@@ -246,12 +246,14 @@ static func get_general_blocks() -> Array[Block]:
 	b.block_format = "Define method {method_name: NIL}"
 	b.statement = "func {method_name}():"
 	b.category = "Communication | Methods"
+	b.tooltip_text = "Define a method/function with following statements"
 	block_list.append(b)
 
 	b = BLOCKS["statement_block"].instantiate()
 	b.block_format = "Call method {method_name: STRING} in group {group: STRING}"
 	b.statement = "get_tree().call_group({group}, {method_name})"
 	b.category = "Communication | Methods"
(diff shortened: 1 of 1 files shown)


manuq (approved):
Does this way of doing tooltips support rich text formatting? In any case I think is a good start, we can improve later.

wjt:
> Does this way of doing tooltips support rich text formatting?

No.

According to https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-property-tooltip-text:

> The default tooltip includes a [PopupPanel](https://docs.godotengine.org/en/stable/classes/class_popuppanel.html#class-popuppanel) and [Label](https://docs.godotengine.org/en/stable/classes/class_label.html#class-label)

Label does not support rich text. If we wanted rich text we would override [_make_custom_tooltip](https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-private-method-make-custom-tooltip) on the root `Block`

manuq:
> > Does this way of doing tooltips support rich text formatting?
> 
> No.
> 
> According to https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-property-tooltip-text:
> 
> > The default tooltip includes a [PopupPanel](https://docs.godotengine.org/en/stable/classes/class_popuppanel.html#class-popuppanel) and [Label](https://docs.godotengine.org/en/stable/classes/class_label.html#class-label)
> 
> Label does not support rich text. If we wanted rich text we would override [_make_custom_tooltip](https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-private-method-make-custom-tooltip) on the root `Block`

OK! I was thinking in potential reuse of the GDScript documentation or put a link to it. Eg. the On Ready block opens the _ready docs. The custom tooltip must be a way to do what the editor tooltips do.
Captura desde 2024-06-27 14-08-30

PR #91, opened 2024-06-27
Fix Simple node setup not happening

wnbaum:
Any Simple node like `SimpleCharacter` will have its `_ready` overwritten by the block script if a ready block exists. This PR enables setup for the Simple node to still happen regardless. (e.g. Pong paddle texture initialization)

This was the best way I could find to do this. Calling `super()` in the child `_ready` when `_ready` is not overridden in the parent causes an error (even though it's a virtual method...). I don't know of any way to check if the method is overridden in the parent to call it or not.

### DIFF
--- a/addons/block_code/block_code_node/block_code.gd
+++ b/addons/block_code/block_code_node/block_code.gd
@@ -60,6 +60,10 @@ func _update_parent_script():
 	for property_name in persist_properties:
 		parent.set(property_name, persist_properties.get(property_name))
 
+	# Run simple setup after node is ready
+	if parent.has_method("simple_setup"):
+		parent.call_deferred("simple_setup")
+
 
 func _get_configuration_warnings():
 	var warnings = []
--- a/addons/block_code/simple_nodes/simple_character/simple_character.gd
+++ b/addons/block_code/simple_nodes/simple_character/simple_character.gd
@@ -19,6 +19,10 @@ func _set_texture(new_texture):
 
 
 func _ready():
+	simple_setup()
+
+
+func simple_setup():
 	_set_texture(texture)
 
 


manuq (commented):
I think it's a good workaround for now. Great!

dbnicholson:
This is an OK workaround, but I feel like we can do better. Would it be possible to extend the generated `_ready` so that it does:

```
diff --git a/addons/block_code/ui/picker/categories/category_factory.gd b/addons/block_code/ui/picker/categories/category_factory.gd
index 1ebb8b1..71659df 100644
--- a/addons/block_code/ui/picker/categories/category_factory.gd
+++ b/addons/block_code/ui/picker/categories/category_factory.gd
@@ -159,7 +159,14 @@ static func get_general_blocks() -> Array[Block]:
        b = BLOCKS["entry_block"].instantiate()
        b.block_name = "ready_block"
        b.block_format = "On Ready"
-       b.statement = "func _ready():"
+       b.statement = (
+               """
+               func _ready():
+                       if get_parent().has_method("_ready"):
+                               super()
+               """
+               . dedent()
+       )
        b.category = "Lifecycle"
        block_list.append(b)
 
```

wnbaum:
Yeah, I would have preferred something like that but from my looking I don't think it's possible. For one, you would want to use `super.has_method("_ready")` instead of `parent.has_method("_ready")` since we are attaching our generated block script to the `BlockCode` node's parent. But when I do `super.has_method("_ready")` it returns true (I'm assuming because it has the method even though it's not implemented). So the issue is there is no way that I could find to determine if the `_ready` method is overridden in the superclass or not. Hence the workaround.

dbnicholson:
Apparently not. If you have `super()` anywhere and it isn't defined in the parent you get a parse error.

PR #92, opened 2024-06-27
Add viewport value blocks

manuq:
To integer blocks for the viewport width and height. And one Vecto2 block for the viewport center.

### DIFF
--- a/addons/block_code/ui/picker/categories/category_factory.gd
+++ b/addons/block_code/ui/picker/categories/category_factory.gd
@@ -490,7 +490,7 @@ static func get_general_blocks() -> Array[Block]:
 	b = BLOCKS["parameter_block"].instantiate()
 	b.variant_type = TYPE_VECTOR2
 	b.block_format = "viewport center"
-	b.statement = "Vector2(%s, %s)" % [viewport_width / 2, viewport_height / 2]
+	b.statement = "get_viewport().get_visible_rect().get_center()"
 	b.category = "Graphics | Viewport"
 	block_list.append(b)
 
--- a/addons/block_code/ui/picker/categories/category_factory.gd
+++ b/addons/block_code/ui/picker/categories/category_factory.gd
@@ -353,6 +353,13 @@ static func get_general_blocks() -> Array[Block]:
 	b.category = "Variables"
 	block_list.append(b)
 
+	b = BLOCKS["parameter_block"].instantiate()
+	b.variant_type = TYPE_VECTOR2
+	b.block_format = "Vector2 x: {x: FLOAT} y: {y: FLOAT}"
+	b.statement = "Vector2({x}, {y})"
+	b.category = "Variables"
+	block_list.append(b)
+
 #endregion
 #region Math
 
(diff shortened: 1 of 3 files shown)


manuq:
Can be used like this:
Captura desde 2024-06-27 16-43-21

wnbaum (commented):
Definitely good to have some sort of blocks like these!

One issue I can think of is that the viewport width, height, and center are all in viewport pixel coordinates, not in our 2D scene coordinates. So when we get the viewport center, it is not the same as the 2D scene center that our camera is fixed on. We might want to add some extra conversion there.

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -462,6 +467,33 @@ static func get_general_blocks() -> Array[Block]:
...
 	block_list.append(b)
+#endregion
+#region Graphics
+
+	var viewport_width = ProjectSettings.get_setting("display/window/size/viewport_width")
+	var viewport_height = ProjectSettings.get_setting("display/window/size/viewport_height")
+
+	b = BLOCKS["parameter_block"].instantiate()
+	b.variant_type = TYPE_INT
+	b.block_format = "viewport width"
```
wnbaum, line 478:
```suggestion
	b.block_format = "Viewport Width"
```
This can probably be caps like the rest of the blocks.

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -462,6 +467,33 @@ static func get_general_blocks() -> Array[Block]:
...
+	b.statement = "%s" % viewport_width
+	b.category = "Graphics | Viewport"
+	block_list.append(b)
+
+	b = BLOCKS["parameter_block"].instantiate()
+	b.variant_type = TYPE_INT
+	b.block_format = "viewport height"
+	b.statement = "%s" % viewport_height
```
wnbaum, line 486: Since the viewport can change size, should we use this as a runtime replacement instead?
```suggestion
	b.statement = "get_viewport().get_visible_rect().size.y"
```

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -462,6 +467,33 @@ static func get_general_blocks() -> Array[Block]:
...
+	b.category = "Graphics | Viewport"
+	block_list.append(b)
+
+	b = BLOCKS["parameter_block"].instantiate()
+	b.variant_type = TYPE_VECTOR2
+	b.block_format = "viewport center"
+	b.statement = "Vector2(%s, %s)" % [viewport_width / 2, viewport_height / 2]
```
wnbaum, line 493:
```suggestion
	b.statement = "get_viewport().get_visible_rect().get_center()"
```

wnbaum (commented):
Here's some suggestions that I forgot to add to the other review!

manuq:
> Definitely good to have some sort of blocks like these!
> 
> One issue I can think of is that the viewport width, height, and center are all in viewport pixel coordinates, not in our 2D scene coordinates. So when we get the viewport center, it is not the same as the 2D scene center that our camera is fixed on. We might want to add some extra conversion there.

Yes, this seems to be the convertion we need: https://forum.godotengine.org/t/how-do-i-get-the-center-of-the-screen-without-using-camera/20545/3?u=manuq

In the current architecture we can't do that easily. I managed to do it with lambda functions. The code looks bad but it works in Pong to reset the ball. See the fixup commit.

I also appended another commit to construct a Vector2 from X, Y float parameters. So these 2 are equivalent:
Captura desde 2024-06-28 10-34-02

wnbaum (approved):
Working as expected! Code looks good to me. The generated code with the lambdas is definitely ugly, but that's something to figure out in another PR.

PR #93, opened 2024-06-27
test_category_factory: Free Block nodes

dbnicholson:
Otherwise the test prints warnings about a pile of orphans.

https://phabricator.endlessm.com/T35520

### DIFF
--- a/tests/test_category_factory.gd
+++ b/tests/test_category_factory.gd
@@ -2,6 +2,13 @@ extends GutTest
 ## Tests for CategoryFactory
 
 
+func free_block_list(blocks: Array[Block]):
+	var block: Block = blocks.pop_back()
+	while block != null:
+		block.free()
+		block = blocks.pop_back()
+
+
 func get_category_names(categories: Array[BlockCategory]) -> Array[String]:
 	var names: Array[String] = []
 	for category in categories:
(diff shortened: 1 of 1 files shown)


wnbaum (approved):
Ah ok nice! I thought for some reason the CategoryFactory was leaking blocks in the plugin, but it was just the test not cleaning up. Looks good to me!

dbnicholson:
> Ah ok nice! I thought for some reason the CategoryFactory was leaking blocks in the plugin, but it was just the test not cleaning up. Looks good to me!

I definitely thought so, too. However, I think that's actually OK because the `Block`s are always transferred into the pickers block list. Later they [get cleaned up](https://github.com/endlessm/godot-block-coding/blob/main/addons/block_code/ui/picker/picker.gd#L35). However, as noted [here](https://phabricator.endlessm.com/T35520#997721), the orphan printing is a feature of GUT. You wouldn't see that in normal operation. You can get the same information from Godot, but it's less clear if we're causing issues or not since it's not nearly as isolated as a single test function. I'm looking at that a bit more. This was just the easy case.

wnbaum:
Yep, as long as there's no nodes that aren't attached to the scene tree I think we're good! I think that when the main panel gets freed on plugin exit, all of the child nodes get freed as well. So as long as we keep all of the blocks we create attached to the picker (or canvas etc.), there should be no leaking.

wjt:
@wnbaum feel free to merge PRs you have positively approved.

PR #95, opened 2024-06-28
ci: Use my fork of godot-asset-lib-action

wjt:
The referenced commit is the merge of these two PRs:

https://github.com/deep-entertainment/godot-asset-lib-action/pull/2
https://github.com/deep-entertainment/godot-asset-lib-action/pull/3


wjt:
The idea here is that I want to test those pull requests – particularly https://github.com/deep-entertainment/godot-asset-lib-action/pull/3 – and the easiest way to do that is to use the merger of those two for our next release of the plugin.

We do not actually *need* either of these changes.

dbnicholson (approved):
Sure, seems reasonable.

PR #96, opened 2024-06-28
Category factory: Expose CharacterBody2D velocity

manuq:
And add a "Move and slide" statement block so it can be used.


manuq:
Usage example:
Captura desde 2024-06-28 11-42-33

wnbaum (approved):
Looks good to me!

PR #98, opened 2024-06-28
Scratch-like category UI

wnbaum:
This adds all uncollapsed block categories back into one large scroll container, but now with buttons that can navigate to each category. This allows the user to see the entire selection of blocks, without the tedium of scrolling. I think this will be nicer for adding in a little menu to create variables within the scrollbox as well.

https://phabricator.endlessm.com/T35532

### DIFF
--- a/addons/block_code/block_code_plugin.gd
+++ b/addons/block_code/block_code_plugin.gd
@@ -41,7 +41,8 @@ const DISABLED_CLASSES := [
 	"Picker",
 	"TitleBar",
 	"MainPanel",
-	"BlockCodePlugin"
+	"BlockCodePlugin",
+	"BlockCategoryButton"
 ]
 
 
--- a/addons/block_code/ui/main_panel.gd
+++ b/addons/block_code/ui/main_panel.gd
@@ -9,11 +9,17 @@ var eia: EditorInterfaceAccess
 @onready var _drag_manager: DragManager = %DragManager
 @onready var _title_bar: TitleBar = %TitleBar
 @onready var _editor_inspector: EditorInspector = EditorInterface.get_inspector()
+@onready var _picker_split: HSplitContainer = %PickerSplit
+@onready var _collapse_button: Button = %CollapseButton
+
+@onready var _icon_collapse := EditorInterface.get_editor_theme().get_icon("Back", "EditorIcons")
+@onready var _icon_expand := EditorInterface.get_editor_theme().get_icon("Forward", "EditorIcons")
 
 var block_code_tab: Button
 var _current_block_code_node: BlockCode
 var _scene_root: Node
 var _block_code_nodes: Array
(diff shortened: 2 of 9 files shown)


dbnicholson:
Looks pretty interesting. Any chance you could attach a couple screenshots?

wnbaum:
image

Here's a screenshot! I just clicked the `Communication | Groups` category.

One thing that could be bothersome is the picker taking up to much space on small screens, but that can be mitigated with zooming and panning in the canvas, and possibly scaling blocks a bit (some of the blocks are very wide too). 

Here's what just the menu looks like in the `Ball` script with a few more categories:

image

wnbaum:
I realized the margins were a bit wrong in some places so I fixed that and rebased to main as well.

manuq:
Looks neat! But yes taking so much horizontal space is a problem. In my screen I can't even see the canvas at all! I would love to keep the picker as a single column. At least if the picker can be resized or even collapsed (like the Scripts panel in the Script editor).

dylanmccall:
I'm definitely a big fan of this layout. To me, it feels much easier to explore, and indeed it gives us a much-needed place for a variables (and signals) UI. But I also agree with that concern about the horizontal space. So, that's a pickle. I suspect this might take some iteration, but I don't want this branch to get lost if we start doing some mockups or something, so I created a Phabricator task to keep track: https://phabricator.endlessm.com/T35532.

One thing that might help is if we can find a way to reduce the size of the blocks in the picker. Or even add some text wrapping, or an optional parameters section that spreads out vertically. (The "Play the sound" block would really benefit from that). It would be useful to noodle around these things anyway, so if we think this minus the five biggest blocks would do the trick, I think it would be okay to merge this in anticipation of further improvements, but let's make a plan for that ;)

manuq:
@wnbaum @dylanmccall [here](https://github.com/endlessm/godot-block-coding/pull/103/commits/34d2972154480659b1134a08b645d98c8e1811ac) I attempted an improvement by replacing some HBoxContainers with HSplitContainers. This works much better in my small screen and more similar to other editor docks.

wnbaum:
@manuq I like the HSplitContainer! I implemented something similar here and also added a little collapse button to get rid of the whole menu if you want. I also added a Ctrl-Space shortcut to easily collapse the Picker which is explained in the tooltip on the collapse button. This shortcut is similar to how Unreal hides the content window in recent versions, I think it's quite nice. Let me know what you think!

wnbaum:
Ok @manuq I made some adjustments, let me know what you think! Not sure about putting the collapse button in the title bar, that seems a little unintuitive; so I pretty much just copied what the normal script editor does. It does sacrifice a bit of vertical space on the canvas, but that space could also be used for error messages and other info in the future.

manuq:
@wnbaum this looks so much better now, very polished!

PR #99, opened 2024-06-29
Add ParameterOutput and monitor node children in SnapPoint

dylanmccall:
This resolves an error that occurred when a block which provides a parameter block was spawning multiple child blocks. Instead of that logic being provided by DragManager, this commit adds a ParameterOutput node which ensures its associated SnapPoint always has a child node, and propagates `drag_started` signals up from the child block to the parent block.

https://phabricator.endlessm.com/T35531


starnight (approved):
LGTM!

PR #104, opened 2024-07-01
Fix getting started instructions

wnbaum:
The getting started instructions imply that you need to clone the repository as a final step to install the plugin which is not the case! Also add a note for those wanting to work in the repo.


dbnicholson (approved):
Yeah, that's better.

PR #105, opened 2024-07-02
plugin.cfg: Use lightweight tags to determine version

wjt:
By default, `git describe` (which is what backs this
`$Format:%(describe)$` placeholder) only uses annotated tags to
construct a version number.

However I have just discovered that if you use the GitHub web interface
to create a tag as part of making a release, the resulting tag is
lightweight, not annotated. As a result the 0.4.0 release describes its
version as `v0.3.1-27-ga9a2ee9`.

Update the placeholder to (the equivalent of) `git describe --tags`,
which considers lightweight tags as well as annotated tags.

### DIFF
--- a/addons/block_code/plugin.cfg
+++ b/addons/block_code/plugin.cfg
@@ -3,5 +3,5 @@
 name="BlockCode"
 description="Create games using a high-level, block-based visual programming language."
 author="Endless"
-version="$Format:%(describe)$"
+version="$Format:%(describe:tags)$"
 script="block_code_plugin.gd"


dbnicholson:
> However I have just discovered that if you use the GitHub web interface to create a tag as part of making a release, the resulting tag is lightweight, not annotated. As a result the 0.4.0 release describes its version as `v0.3.1-27-ga9a2ee9`.

Wow, that's unfortunate. Apparently we're [not the only ones to notice that](https://github.com/orgs/community/discussions/4924). One thing we could do is use our own automation for the tags and release. We did that in kolibri-explore-plugin. Or just create and push the tags locally and then create the release from an existing tag instead of creating a new one.

wjt:
I did a little bit of experimenting and if we were feeling really keen we could make the workflow that is run for releases check that the tag is annotated:

https://github.com/endlessm/godot-block-coding/pull/108

PR #106, opened 2024-07-02
ui: Use EditorInterface method to display Block Code editor

dbnicholson:
Rather than locating the Block Code tab in the editor tab bar and simulating a button press on it, just use the
EditorInterface.set_main_screen_editor method to display the Block Code editor. With that, the EditorInterfaceAccess class (and utilities) vendored from GDTour can be dropped.


dbnicholson:
I'm pretty sure this has the exact same effect but I don't totally understand how the UI works.

wnbaum (approved):
Looks good to me! The `EditorInterfaceAccess` script isn't really useful anymore, unless we want to modify the editor UI. We can always add it back if we need to, but I don't really see a use for it in the near future. Good find with `set_main_screen_editor()`!

PR #107, opened 2024-07-02
Revamp variable system

wnbaum:
This removes the old variable system we had with the VAR_DICT, and replaces it with a menu for creating typed variables that you can use as blocks.

image

Side note: We will need to fix the sound blocks after this is merged, since they use the VAR_DICT. I think it would be good to have some library functions or a singleton our blocks can call on for more behavior like this.

https://phabricator.endlessm.com/T35535

### DIFF
--- a/addons/block_code/types/types.gd
+++ b/addons/block_code/types/types.gd
@@ -36,6 +36,7 @@ const cast_relationships = [
 	[TYPE_INT, TYPE_STRING, "str(%s)"],
 	[TYPE_FLOAT, TYPE_STRING, "str(%s)"],
 	[TYPE_COLOR, TYPE_STRING, "str(%s)"],
+	[TYPE_VECTOR2, TYPE_STRING, "str(%s)"],
 ]
 
 # Directed graph, edges are CastGraphEdge
--- a/addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
@@ -46,6 +46,8 @@ func _generate_script_from_entry_blocks(entry_statement: String, entry_blocks: A
 	var signal_node: InstructionTree.TreeNode
 	var is_empty = true
 
+	InstructionTree.IDHandler.reset()
+
 	for entry_block in entry_blocks:
 		var next_block := entry_block.bottom_snap.get_snapped_block()
 
--- a/addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
@@ -24,7 +24,7 @@ func generate_script_from_current_window(bsd: BlockScriptData):
 	script += "extends %s\n\n" % bsd.script_inherits
 
 	for variable in bsd.variables:
-		script += "var %s\n\n" % variable.var_name
+		script += "var %s: %s\n\n" % [variable.var_name, type_string(variable.var_type)]
 
 	script += "\n"
 
--- a/addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
@@ -3,7 +3,7 @@ class_name NodeBlockCanvas
 extends BlockCanvas
 
 
-func generate_script_from_current_window(script_inherits: String = ""):
+func generate_script_from_current_window(bsd: BlockScriptData):
 	# TODO: implement multiple windows
 	var current_window := _window
 
@@ -21,9 +21,12 @@ func generate_script_from_current_window(script_inherits: String = ""):
 
 	var script: String = ""
 
-	script += "extends %s\n\n" % script_inherits
+	script += "extends %s\n\n" % bsd.script_inherits
 
-	script += "var VAR_DICT := {}\n\n"
+	for variable in bsd.variables:
+		script += "var %s\n\n" % variable.var_name
+
+	script += "\n"
 
 	var init_func = InstructionTree.TreeNode.new("func _init():")
 
--- a/addons/block_code/ui/picker/categories/category_factory.gd
+++ b/addons/block_code/ui/picker/categories/category_factory.gd
@@ -204,7 +204,7 @@ static func get_general_blocks() -> Array[Block]:
 
 	b = BLOCKS["control_block"].instantiate()
 	b.block_formats = ["repeat {number: INT}"]
-	b.statements = ["for i in {number}:"]
+	b.statements = ["for __i in {number}:"]
 	b.category = "Loops"
 	block_list.append(b)
 
@@ -423,10 +423,10 @@ static func get_general_blocks() -> Array[Block]:
 	b.block_format = "Load file {file_path: STRING} as sound {name: STRING}"
 	b.statement = (
 		"""
-		VAR_DICT[{name}] = AudioStreamPlayer.new()
-		VAR_DICT[{name}].name = {name}
-		VAR_DICT[{name}].set_stream(load({file_path}))
-		add_child(VAR_DICT[{name}])
+		var __sound = AudioStreamPlayer.new()
+		__sound.name = {name}
+		__sound.set_stream(load({file_path}))
+		add_child(__sound)
 		"""
 		. dedent()
 	)
@@ -439,9 +439,10 @@ static func get_general_blocks() -> Array[Block]:
 	b.block_format = "Play the sound {name: STRING} with Volume dB {db: FLOAT} and Pitch Scale {pitch: FLOAT}"
 	b.statement = (
 		"""
-		VAR_DICT[{name}].volume_db = {db}
-		VAR_DICT[{name}].pitch_scale = {pitch}
-		VAR_DICT[{name}].play()
+		var __sound_node = get_node({name})
+		__sound_node.volume_db = {db}
+		__sound_node.pitch_scale = {pitch}
+		__sound_node.play()
 		"""
 		. dedent()
 	)
--- a/addons/block_code/ui/picker/categories/category_factory.gd
+++ b/addons/block_code/ui/picker/categories/category_factory.gd
@@ -328,38 +328,6 @@ static func get_general_blocks() -> Array[Block]:
 
 #endregion
 #region Variables
-
-	b = BLOCKS["statement_block"].instantiate()
-	b.block_format = "Set String {var: STRING} {value: STRING}"
-	b.statement = "VAR_DICT[{var}] = {value}"
-	b.category = "Variables"
-	block_list.append(b)
-
-	b = BLOCKS["parameter_block"].instantiate()
-	b.block_format = "Get String {var: STRING}"
-	b.statement = "VAR_DICT[{var}]"
-	b.category = "Variables"
-	block_list.append(b)
-
-	b = BLOCKS["statement_block"].instantiate()
-	b.block_format = "Set Int {var: STRING} {value: INT}"
-	b.statement = "VAR_DICT[{var}] = {value}"
-	b.category = "Variables"
-	block_list.append(b)
-
-	b = BLOCKS["parameter_block"].instantiate()
-	b.variant_type = TYPE_INT
-	b.block_format = "Get Int {var: STRING}"
-	b.statement = "VAR_DICT[{var}]"
-	b.category = "Variables"
-	block_list.append(b)
-
-	b = BLOCKS["parameter_block"].instantiate()
-	b.block_format = "To String {int: INT}"
-	b.statement = "str({int})"
-	b.category = "Variables"
-	block_list.append(b)
-
 	b = BLOCKS["parameter_block"].instantiate()
 	b.variant_type = TYPE_VECTOR2
 	b.block_format = "Vector2 x: {x: FLOAT} y: {y: FLOAT}"
@@ -513,10 +481,10 @@ static func get_general_blocks() -> Array[Block]:
 static func property_to_blocklist(property: Dictionary) -> Array[Block]:
 	var block_list: Array[Block] = []
 
-	var block_type = property.type
+	var variant_type = property.type
 
-	if block_type:
-		var type_string: String = Types.VARIANT_TYPE_TO_STRING[block_type]
+	if variant_type:
+		var type_string: String = Types.VARIANT_TYPE_TO_STRING[variant_type]
 
 		var b = BLOCKS["statement_block"].instantiate()
 		b.block_format = "Set %s to {value: %s}" % [property.name.capitalize(), type_string]
@@ -531,7 +499,7 @@ static func property_to_blocklist(property: Dictionary) -> Array[Block]:
 		block_list.append(b)
 
 		b = BLOCKS["parameter_block"].instantiate()
-		b.block_type = block_type
+		b.variant_type = variant_type
 		b.block_format = "%s" % property.name.capitalize()
 		b.statement = "%s" % property.name
 		b.category = property.category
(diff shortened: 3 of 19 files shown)


addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
```diff
@@ -21,9 +21,12 @@ func generate_script_from_current_window(script_inherits: String = ""):
 
 	var script: String = ""
 
-	script += "extends %s\n\n" % script_inherits
+	script += "extends %s\n\n" % bsd.script_inherits
 
-	script += "var VAR_DICT := {}\n\n"
+	for variable in bsd.variables:
+		script += "var %s\n\n" % variable.var_name.to_snake_case()
```
dbnicholson, line 27: Is it possible to declare these as typed variables? One advantage of typed variables is that they're [initialized to a default] instead of `null`.

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -320,32 +320,6 @@ static func get_general_blocks() -> Array[Block]:
 
 #endregion
 #region Variables
-
-	b = BLOCKS["statement_block"].instantiate()
-	b.block_format = "Set String {var: STRING} {value: STRING}"
-	b.statement = "VAR_DICT[{var}] = {value}"
```
dbnicholson, line 326: Will removing support for `VAR_DICT` break existing serializations? Do we need some kind of migration routine to handle it?

addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
wnbaum: Yea, it would be nice to have defaults in there somehow, I was thinking about how we would accomplish that. At the moment we have no way to convert from a `Variant.Type` to the text Godot uses to specify types, but we could definitely have another dictionary in `types.gd` that maps from a `Variant.Type` to a default value. I'll take a look at that.

addons/block_code/ui/picker/categories/category_factory.gd
wnbaum: The way we are serialize blocks makes it so that (unfortunately) we can't modify blocks once they are placed in the scene. [T35536](https://phabricator.endlessm.com/T35536) will hopefully fix this.

We could have a migration routine, but I think it's probably best we just spend time finishing that ticket so we don't have to worry about making similar routines in the future.

This PR does break the sound blocks though... should I append a commit here to fix them? It's a rather large change so I could make another PR and just disable them here temporarily.

dbnicholson: I think the sound blocks is kinda the point, right? If it's happening in our blocks it's certainly going to happen for others. What happens when you have a scene with a block script and a variable of some kind and then you upgrade the plugin?

I'd like to see the commit to see what's going to be required. It's early enough where we could say "sorry, your script might be broken", but at least I think we need to know what you'd have to do to fix it.

wnbaum: Yeah, when you upgrade the plugin, the blocks in each script don't get upgraded, you have to manually go in and replace each of them if you want the new ones from the picker... which is not ideal. So if anyone has used a `VAR_DICT` style variable in their scenes, it will break.

This has happened a few times in the past when we have upgraded blocks, and even the pong game example is pretty outdated, though the old blocks still work. 

I suppose we could throw in something to the code generator that detects if a `VAR_DICT` statement is in the script and adds it back to the top for legacy scripts. I guess I just don't want the old `VAR_DICT` stuff to haunt me in the current code, but I can see why it's good to have support for older versions 😄

dbnicholson: I don't think you'd try to keep the `VAR_DICT` statements. I think you'd detect them and rewrite them using the new syntax. I don't know how difficult that would be, but it's something I think we're going to end up dealing with more than once.

The serialization stored in the project's resource files are essentially ABI. If we break it, we're going to annoy our users. So, we need to get it in the shape we want ASAP, but we also need to be honest that we're going to find issues with it (like we are now) and that we have to have a plan for managing those transitions.

There will always be problems that come up that affect the serialization, whether we want it to or not. I'm not saying you personally need to come up with the plan to handle it. It's something the whole team needs to think about and be aware of as we continue iterating. To say it another way, the serialization is maybe the most critical part of the whole plugin because changes in it immediately break users' projects. Nothing could be more annoying than updating our plugin and finding out your project is now broken.

addons/block_code/types/types.gd
```diff
@@ -30,6 +30,16 @@ const STRING_TO_VARIANT_TYPE: Dictionary = {
...
 }
 
+const VARIANT_TYPE_TO_GDSCRIPT_STRING: Dictionary = {
+	TYPE_STRING: "String",
+	TYPE_INT: "int",
+	TYPE_FLOAT: "float",
+	TYPE_BOOL: "bool",
+	TYPE_VECTOR2: "Vector2",
+	TYPE_COLOR: "Color",
+	TYPE_NODE_PATH: "NodePath",
```
manuq, line 40: YES!!

addons/block_code/ui/picker/categories/category_factory.gd
dylanmccall: > I don't think you'd try to keep the VAR_DICT statements. I think you'd detect them and rewrite them using the new syntax. I don't know how difficult that would be, but it's something I think we're going to end up dealing with more than once.

Okay, I created https://phabricator.endlessm.com/T35547 about adding a migration system. Indeed, it's a good thing to do early before we really need it :) I don't think we need to block on adding migrations for this particular case this early in the thing's life, but then it _could_ be a useful test case.

dbnicholson: After talking to Daniel, he suggested it was too early to worry about backwards compatibility since we haven't even had this in front of learners yet. So, we can just ignore this for now, but it's definitely something to keep in mind going forward.

addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
dbnicholson: Doesn't [type_string](https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#class-globalscope-method-type-string) do that?

dbnicholson:
After punting on the compatibility, I still have a few questions.

1. Can you add the commit that manually migrates the sound blocks away from `VAR_DICT`? I think that's useful to see what the actual effect of this change is.
2. Is this blocked on https://phabricator.endlessm.com/T35536 or do you want this to go in now?
3. Have you discussed the UI changes with the learning team? I think there's no doubt the menu/dialog is an improvement, but it would be useful to have some design input.

addons/block_code/types/types.gd
dbnicholson: Isn't this just duplicating [type_string](https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#class-globalscope-method-type-string)? The following test passed for me:

```
extends GutTest

const VARIANT_TYPE_TO_GDSCRIPT_STRING: Dictionary = {
	TYPE_STRING: "String",
	TYPE_INT: "int",
	TYPE_FLOAT: "float",
	TYPE_BOOL: "bool",
	TYPE_VECTOR2: "Vector2",
	TYPE_COLOR: "Color",
	TYPE_NODE_PATH: "NodePath",
}

func test_type_string():
	for type in VARIANT_TYPE_TO_GDSCRIPT_STRING:
		var expected: String = VARIANT_TYPE_TO_GDSCRIPT_STRING[type]
		var type_str: String = type_string(type)
		assert_eq(type_str, expected)
```

wnbaum: Correct! I don't know how I didn't think of that.

wnbaum:
>Can you add the commit that manually migrates the sound blocks away from VAR_DICT? I think that's useful to see what the actual effect of this change is.

Sure, the sound blocks using `VAR_DICT` felt kinda hacky to me when it was added, so it'll be good to fix that.

>Is this blocked on https://phabricator.endlessm.com/T35536 or do you want this to go in now?

Definitely not blocked, only really needed T35536 if we were going to make a migration system, which we will hold off on for now.

>Have you discussed the UI changes with the learning team? I think there's no doubt the menu/dialog is an improvement, but it would be useful to have some design input.

I haven't! Will definitely get some feedback and fix the sound blocks.

wnbaum:
Converted to draft for now as I work on fixing the sound blocks, which has brought up more of a problem than I expected. Will continue tomorrow, but push my working but ugly solution now.

wnbaum:
Ok, after fixing some weird bugs, it seems to be working great.

New variables system is working as before.

You can now define unique local variables in block statements by suffixing them like `var test__`, which will give them an ID in the generated code like `var test_1`. This fixes the sound blocks, and also same-name for loop index variables. VAR_DICT is no more.

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -766,3 +740,29 @@ static func _get_input_blocks() -> Array[Block]:
...
+
+	for variable in variables:
+		var capital_name: String = variable.var_name.capitalize()
+		var camel_name: String = variable.var_name.to_snake_case()
+		var type_string: String = Types.VARIANT_TYPE_TO_STRING[variable.var_type]
+
+		var b = BLOCKS["parameter_block"].instantiate()
+		b.variant_type = variable.var_type
+		b.block_format = capital_name
+		b.statement = camel_name
```
dbnicholson, line 757: It's a bit strange that the entered, displayed and formatted variables are all different. Why not just use the variable name as was entered by the user? What happens when the user creates `foo` and `Foo` variables? Legally these are two different variables, but you won't be able to distinguish them in the UI and they'll conflict in the generated script.

dbnicholson (changes requested):
Overall, I like this a lot. Thanks for changing the sound blocks in pong, too. That shows the practical implication of this kind of change.

addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
```diff
@@ -51,6 +51,7 @@ func _generate_script_from_entry_blocks(entry_statement: String, entry_blocks: A
 
 		if next_block != null:
 			var generator: InstructionTree = InstructionTree.new()
+			InstructionTree.IDHandler.reset()
```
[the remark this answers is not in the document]
wnbaum: Well, any top level variables are user defined so they should be different from each other. But I realize from your other comment that I didn't really check that anywhere.

And you're right, I didn't really want to pass the instruction tree into `get_instruction_node()` but I guess the generated code is dependant on the state of the instruction tree so it's fine. I'll keep track of the counts non-statically. I think we'll change the code generation once we integrate the new block resources anyways.

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -204,7 +204,7 @@ static func get_general_blocks() -> Array[Block]:
 
 	b = BLOCKS["control_block"].instantiate()
 	b.block_formats = ["repeat {number: INT}"]
-	b.statements = ["for i in {number}:"]
+	b.statements = ["for i__ in {number}:"]
```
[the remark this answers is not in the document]
wnbaum: Yup, needed for nested loops. If I pass the instruction tree into the `get_instruction_node()` method, it might be easier to handle the going out of scope in which case you could use the same variable names. I just tried to get it working in the least amount of changes, since we'll likely have to rewrite it for the new blocks anyways.

addons/block_code/ui/picker/categories/category_factory.gd
wnbaum: Yea my idea was to make them look pretty but it's probably better to just have the display and formatted be the same as what the user types. Will also have to check for invalid/duplicate names.

wnbaum:
Thanks for the review @dbnicholson there's definitely some changes I should make here.

dylanmccall (commented):
I'm happy with this. A couple small suggestions for things to tweak. It bothered me that `variables` is a property of BlockScriptData instead of part of block_trees, but I see how it would be difficult to do the latter with our current design.

Some UX things which I can understand leaving for a follow-up, but I'll note them here for the moment. In this screenshot…

Screenshot from 2024-07-12 14-11-28

- Because of the notch at the bottom, the "Set Name to …" block looks like it's closer to "Age" (below it) than to "Name" (above it).
- I'm finding all this white on orange text kind of difficult to parse. Two reasons: shallow colour contrast, and it's just a lot of orange. We definitely need to have a good discussion about block design in general at some point.
- You didn't touch it here, but I'm not sure if that "To String" block is doing anything for us.

addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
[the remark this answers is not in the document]
dbnicholson: Right, probably good enough as is. However, maybe I'm misunderstanding how entry blocks are handled, but shouldn't this be reset once per entry statement to be scoped correctly? For example, suppose I have multiple `ready_block`s. `generate_script_from_current_window` will collect them into an array keyed by the `func _ready():` so that they can be serialized into a single function body. Within that function, we would want to reset the ID counter once so the names are unique, not once per entry block.

If I'm understanding correctly, you could put 2 `On Ready` blocks on the canvas and have them each load a sound. The sound variables would end up both being named `sound_0` and conflict. Am I reading that correctly?

addons/block_code/ui/picker/categories/category_factory.gd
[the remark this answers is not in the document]
dbnicholson: Yeah, good enough for now. I just wanted to make sure I was understanding correctly.

addons/block_code/ui/block_canvas/node_block_canvas/node_block_canvas.gd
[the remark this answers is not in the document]
wnbaum: Oh, you're right. For some reason I thought the `_generate_script_from_entry_blocks` was referring to all of the entry blocks, not the ones with the same names. Resetting the counter at the beginning of `_generate_script_from_entry_blocks` before the loop should work properly.

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -455,10 +429,10 @@ static func get_general_blocks() -> Array[Block]:
 	b.block_format = "Load file {file_path: STRING} as sound {name: STRING}"
 	b.statement = (
 		"""
-		VAR_DICT[{name}] = AudioStreamPlayer.new()
-		VAR_DICT[{name}].name = {name}
-		VAR_DICT[{name}].set_stream(load({file_path}))
-		add_child(VAR_DICT[{name}])
+		var sound__ = AudioStreamPlayer.new()
+		sound__.name = {name}
```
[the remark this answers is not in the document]
dbnicholson: I think that's a good idea even though with the suffixed counter it's unlikely to clash. Still, I think the common convention is that you _prefix_ internal variables with `_` or `__`. Furthermore, we could add in something semi-unique like `bcp` (for Block Code Plugin). So, `__bcp_<something>`.

In fact, if this variable isn't going to be exposed to the user, then there's no need to name it like the thing it's for at all. So, you just declare all the internal variables `__bcp`. The generator recognizes that (without tricky regex matching or trimming any parts) and suffixes the generated variable with a `<counter>` or `_<counter>`.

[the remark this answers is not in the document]
wnbaum: I think the whole point of generating the script like this is to have it exposed to the user, otherwise we could just have a library function like `Lib.play_sound()` that would bypass the need for local variables anyways.

That's also why I implemented a counter, instead of having a unique ID appended to each variable, because it makes the code look a lot nicer. Anyways, a prefix will definitely help here since we can also prevent the user from creating variables with our prefix.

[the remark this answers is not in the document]
dbnicholson: That's fair. I think just `_` or `__` prefix is good enough to make conflict with a user supplied variable extremely unlikely.

wnbaum:
Ok, I made a bunch of requested changes. Ready for review!

1. Local variable ID counter is reset at the beginning of each entry block
2. Variables now use the raw name the user puts in and are checked for validity/conflicts with existing variables and possible local variables
3. Local variables are now prefixed with `__` rather than suffixed
4. To String Block removed
5. Spacer added between groups of variable blocks in the picker for clarity
6. Variable create dialog scene separated from variable create button scene

dylanmccall (approved):

dbnicholson (approved):
Admittedly I kinda glossed over the last set of changes, but it looks like you addressed the things that I commented on. Time to land this plane!

PR #110, opened 2024-07-03
Simplify background polygon

wjt:
A couple of simplifications to building the shape of the background polygon.

I was looking at this code with an eye to fixing the left border of `if` blocks, but did not get to that yet. https://phabricator.endlessm.com/T35538

### DIFF
--- a/addons/block_code/ui/blocks/utilities/background/background.gd
+++ b/addons/block_code/ui/blocks/utilities/background/background.gd
@@ -46,65 +46,58 @@ func _set_shift_bottom(new_shift_bottom):
 	queue_redraw()
 
 
-func float_array_to_Vector2Array(coords: Array) -> PackedVector2Array:
-	# Convert the array of floats into a PackedVector2Array.
-	var array: PackedVector2Array = []
-	for coord in coords:
-		array.append(Vector2(coord[0], coord[1]))
-	return array
-
-
 func _draw():
 	outline_color = color.darkened(0.2)
 
-	var fill_polygon = [[0.0, 0.0]]
+	var fill_polygon: PackedVector2Array
+	fill_polygon.append(Vector2(0.0, 0.0))
 	if show_top:
-		fill_polygon += [
-			[Constants.KNOB_X + shift_top, 0.0],
-			[Constants.KNOB_X + Constants.KNOB_Z + shift_top, Constants.KNOB_H],
-			[Constants.KNOB_X + Constants.KNOB_Z + Constants.KNOB_W + shift_top, Constants.KNOB_H],
-			[Constants.KNOB_X + Constants.KNOB_Z * 2 + Constants.KNOB_W + shift_top, 0.0],
+		fill_polygon.append_array(PackedVector2Array(
+			[
+				Vector2(Constants.KNOB_X + shift_top, 0.0),
+				Vector2(Constants.KNOB_X + Constants.KNOB_Z + shift_top, Constants.KNOB_H),
+				Vector2(Constants.KNOB_X + Constants.KNOB_Z + Constants.KNOB_W + shift_top, Constants.KNOB_H),
+				Vector2(Constants.KNOB_X + Constants.KNOB_Z * 2 + Constants.KNOB_W + shift_top, 0.0),
+			]
+		))
+	fill_polygon.append_array(PackedVector2Array(
+		[
+			Vector2(size.x, 0.0),
+			Vector2(size.x, size.y),
+			Vector2(Constants.KNOB_X + Constants.KNOB_Z * 2 + Constants.KNOB_W + shift_bottom, size.y),
+			Vector2(Constants.KNOB_X + Constants.KNOB_Z + Constants.KNOB_W + shift_bottom, size.y + Constants.KNOB_H),
+			Vector2(Constants.KNOB_X + Constants.KNOB_Z + shift_bottom, size.y + Constants.KNOB_H),
+			Vector2(Constants.KNOB_X + shift_bottom, size.y),
+			Vector2(0.0, size.y),
+			Vector2(0.0, 0.0),
 		]
-	fill_polygon += [
-		[size.x, 0.0],
-		[size.x, size.y],
-		[Constants.KNOB_X + Constants.KNOB_Z * 2 + Constants.KNOB_W + shift_bottom, size.y],
-		[Constants.KNOB_X + Constants.KNOB_Z + Constants.KNOB_W + shift_bottom, size.y + Constants.KNOB_H],
-		[Constants.KNOB_X + Constants.KNOB_Z + shift_bottom, size.y + Constants.KNOB_H],
-		[Constants.KNOB_X + shift_bottom, size.y],
-		[0.0, size.y],
-		[0.0, 0.0],
-	]
+	))
 
-	var stroke_polygon = []
-	stroke_polygon += [
-		[shift_top, 0.0],
-	]
+	var stroke_polygon: PackedVector2Array
+	stroke_polygon.append(Vector2(shift_top, 0.0))
 	if show_top:
-		stroke_polygon += [
-			[Constants.KNOB_X + shift_top, 0.0],
-			[Constants.KNOB_X + Constants.KNOB_Z + shift_top, Constants.KNOB_H],
-			[Constants.KNOB_X + Constants.KNOB_Z + Constants.KNOB_W + shift_top, Constants.KNOB_H],
-			[Constants.KNOB_X + Constants.KNOB_Z * 2 + Constants.KNOB_W + shift_top, 0.0],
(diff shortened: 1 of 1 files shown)


addons/block_code/ui/blocks/utilities/background/background.gd
```diff
@@ -46,74 +46,58 @@ func _set_shift_bottom(new_shift_bottom):
...
-	for coord in coords:
-		array.append(Vector2(coord[0], coord[1]))
-	return array
-
-
 func _draw():
 	outline_color = color.darkened(0.2)
 
-	var fill_polygon = [[0.0, 0.0]]
+	var fill_polygon: PackedVector2Array
```
dbnicholson, line 52: I'd think you'd need to initialize this with a constructor like `PackedVector2Array()` or `[]`.

addons/block_code/ui/blocks/utilities/background/background.gd
```diff
@@ -46,74 +46,58 @@ func _set_shift_bottom(new_shift_bottom):
...
-
-	var stroke_polygon = []
-	if shift_top == 0:
-		stroke_polygon += [
-			[0.0, 0.0],
-		]
-	else:
-		stroke_polygon += [
-			[shift_top, 0.0],
+		fill_polygon += PackedVector2Array(
```
dbnicholson, line 55: I don't know exactly how `+=` composes, but according to the [documentation](https://docs.godotengine.org/en/stable/classes/class_packedvector2array.html#class-packedvector2array-operator-sum-packedvector2array) `+` returns a new array and `append_array` should be preferred for performance.

dbnicholson (changes requested):
This is a nice fix I was also looking at locally. A bigger and better change would be to cache the polygons instead of always computing them in the `_draw` handler since they only change when the various exported variables or `size` changes.

wjt:
> A bigger and better change would be to cache the polygons instead of always computing them in the `_draw` handler since they only change when the various exported variables or `size` changes.

Great point.

I'll look at the other adjustments you suggest above.

addons/block_code/ui/blocks/utilities/background/background.gd
wjt: One might think that but actually you don't need to.

wjt:
> > A bigger and better change would be to cache the polygons instead of always computing them in the `_draw` handler since they only change when the various exported variables or `size` changes.

I experimented with this and actually as far as I can tell `_draw` is only called in response to the `queue_redraw()` calls so there is no upside to caching the polygons.

addons/block_code/ui/blocks/utilities/background/background.gd
wjt: I read the interpreter source code a little. It does look like `x += y` is implemented as:

```
tmp = x + y
x = tmp
```

However if I change all these calls to `append_array` then gdformat suggests the following completely unhinged formatting:

```diff
 	if show_top:
-		fill_polygon.append_array(PackedVector2Array(
-			[
-				Vector2(Constants.KNOB_X + shift_top, 0.0),
-				Vector2(Constants.KNOB_X + Constants.KNOB_Z + shift_top, Constants.KNOB_H),
-				Vector2(Constants.KNOB_X + Constants.KNOB_Z + Constants.KNOB_W + shift_top, Constants.KNOB_H),
-				Vector2(Constants.KNOB_X + Constants.KNOB_Z * 2 + Constants.KNOB_W + shift_top, 0.0),
-			]
-		))
+		(
+			fill_polygon
+			. append_array(
+				PackedVector2Array(
+					[
+						Vector2(Constants.KNOB_X + shift_top, 0.0),
+						Vector2(Constants.KNOB_X + Constants.KNOB_Z + shift_top, Constants.KNOB_H),
+						Vector2(Constants.KNOB_X + Constants.KNOB_Z + Constants.KNOB_W + shift_top, Constants.KNOB_H),
+						Vector2(Constants.KNOB_X + Constants.KNOB_Z * 2 + Constants.KNOB_W + shift_top, 0.0),
+					]
+				)
+			)
+		)
```

I'm all for deferring formatting suggestions to a tool but this is a really bad suggestion.

addons/block_code/ui/blocks/utilities/background/background.gd
```diff
@@ -46,74 +46,44 @@ func _set_shift_bottom(new_shift_bottom):
...
-	var stroke_polygon = []
-	if shift_top == 0:
-		stroke_polygon += [
-			[0.0, 0.0],
-		]
-	else:
-		stroke_polygon += [
-			[shift_top, 0.0],
-		]
+		fill_polygon.append(Vector2(Constants.KNOB_X + shift_top, 0.0))
```
wjt, line 55: I'm really not wild about the repeated:

```
fill_polygon.append(Vector2(...))
```

noise. I almost think that the original code is preferable.

addons/block_code/ui/blocks/utilities/background/background.gd
dbnicholson: I guess typed variables are automatically initialized to a default value.

https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#initialization-order

addons/block_code/ui/blocks/utilities/background/background.gd
dbnicholson: That's horrifying but personally I'd just let it do it if it's required for better code.

dbnicholson (approved):
I leave it to you to decide the formatting vs original code. One thing that could split the difference and may be more efficient is to build a regular array with `Vector2` and then convert to a packed array using the constructor at the end.

addons/block_code/ui/blocks/utilities/background/background.gd
wjt: I doubt that the `x.append_array(PackedVector2Array([a, b, c]))` syntax is significantly faster than `x += PackedVector2Array([a, b, c])`. Either way we are constructing an unpacked array, then constructing a packed array, doing a thing, and discarding both temporaries. What's a third temporary between friends.

wjt: I'm totally wrong!

I profiled three different ways to append 100,000 `Vector2.ZERO`s to a PackedVector2Array.

```gdscript
func append_one():
	var x: PackedVector2Array
	for _i in range(100_000):
		x.append(Vector2.ZERO)
	return x.size()

func append_array():
	var x: PackedVector2Array
	for _i in range(100_000):
		x.append_array(PackedVector2Array([Vector2.ZERO]))
	return x.size()

func plus_equals():
	var x: PackedVector2Array
	for _i in range(100_000):
		x += PackedVector2Array([Vector2.ZERO])
	return x.size()
```

Taking the mean of 10 trials in each case:

- append_one: 12ms
- append_array: 51ms
- plus_equals: 2342ms

wjt: I'll stick with the repeated calls to `.append()` because the horrifying formatting from gdformat makes the conditional blocks' indentation actively hard to follow.

wjt:
I decided to go with the wordy code that allocates the fewest temporaries.

wjt:
Here is the existing bug report for the bad gdformat output: https://github.com/Scony/godot-gdscript-toolkit/issues/295

addons/block_code/ui/blocks/utilities/background/background.gd
dbnicholson: Interesting analysis, but is that a fair comparison? The point of using `append_array` or `+=` is to append multiple values in one call. You wouldn't use it to add a single value array. It would be more interesting to see the entire range added in one call.

wjt: https://gist.github.com/wjt/0aae0ed3d9f0bd82991f3d17e0e5d608 appends the same 80,008-element `PackedVector2Array` to a new array, 100 times.

* Node2D(game.gd)::append_array: 80ms
* Node2D(game.gd)::plus_equals: 2233ms

wjt: Without doing it 100 times they are both instant. Ultimately we're working with such little data that it probably makes zero difference!

dbnicholson: Indeed, I'm sure it makes no difference in this case! It's more morbid curiosity. What I had in mind was more like https://gist.github.com/dbnicholson/5ab252bf748944ed37874ac7f9268e87 where a few values at a time are being added. I wanted to see the difference between adding a single element vs an array of several elements.

Node(test.gd)::append_one: 20ms
Node(test.gd)::append_array: 6ms
Node(test.gd)::append_array_one: 118ms

I didn't bother with `+=` since you already verified the documentation and it's ridiculously slower than either append method.

PR #114, opened 2024-07-03
Use is_physical_key_pressed for game input

wjt:
The documentation of `Input.is_physical_key_pressed` says:

> `is_physical_key_pressed()` is recommended over `is_key_pressed()` for
> in-game actions, as it will make `W`/`A`/`S`/`D` layouts work regardless of
> the user's keyboard layout. `is_physical_key_pressed()` will also ensure that
> the top row number keys work on any keyboard layout. If in doubt, use
> `is_physical_key_pressed()`.

The physical keys labelled `W` `A` `S` `D` on a US keyboard enter different characters on some non-US keyboard layouts:

- AZERTY (standard in France):  `Z` `Q` `S` `D`
- Dvorak (standard on my desk): `,` `A` `O` `E`

(The other widely-used layout for Latin text is QWERTZ, used primarily in German-speaking regions; though I have learned while trying to justify this change that the Serbian Cyrillic layout is a rare example of a non-Latin layout that is based on QWERTZ, presumably for geographic reasons. But WASD are the same in QWERTZ and QWERTY.)

Adjust the “Move with player 1/2 buttons” blocks accordingly.

I have not adjusted the “Move with keys {up: STRING} {down: STRING} {left: STRING} {right: STRING} with speed {speed: VECTOR2}” block because when the user enters strings into that block, they will be entering the logical character corresponding to the physical key on their keyboard that they want to use. It would be very confusing if a French user configured that block to use keys Z S Q D and then those keys were mapped to their physical locations on a US keyboard. The distinction between physical keys and logical keys is confusing and not something we want to try to explain here (IMO).


dbnicholson (approved):
Dvorak strikes again! :bow_and_arrow:

PR #115, opened 2024-07-03
background: Avoid infinite draw cycle

dbnicholson:
Setting outline_color queues a redraw, so setting it in the draw handler implies that it will queue infinite redraws. In Godot 4.2 this doesn't happen, but in Godot 4.3 it does and causes the editor to lock up until it finally crashes.

I don't know if it's a bug in one version or the other, but either way we don't need to do that. If outline_color is already set based on the color, then set it at the time color is set and stop exporting it since any user provided value would be ignored.


manuq (approved):
Oh, it was that! Genius.

PR #117, opened 2024-07-03
T35541 block code in bottom panel

dylanmccall:
https://phabricator.endlessm.com/T35541


dylanmccall:
This is _technically_ ready to review, but I'm marking it as a draft because we need additional feedback related to the design.

The big change here is removing the block code editor as a main panel, and instead adding it to the bottom bar:

Screenshot from 2024-07-03 16-49-38

As before, it appears automatically if a `BlockCode` node is selected. In addition, it appears if a node _containing_ a `BlockCode` node is selected. That is useful, because now we can work with the Block Code editor while manipulating objects in the canvas! For example, the Ball in the scene is highlighted because I clicked it, and the block code editor shows the code associated with it.

The editor is very, very busy here, but fortunately Godot does have a button on the bottom right corner to expand the bottom panel, which results in a similar experience to what we have right now:

Screenshot from 2024-07-03 16-52-33

With Block Code in the bottom panel, we gain another neat trick: we can be cleverer about empty states. In particular, since we are working in the same context as the rest of the scene, the Block Code editor can provide helper text if the scene is empty, and it can guide the user to creating a `BlockCode` node when they have selected a node in the scene. For example, here I have selected the `Camera2D` node in the scene, and the editor displays a convenient button to add block code for that node:

Screenshot from 2024-07-03 16-56-18

manuq:
@dylanmccall this is a huge improvement! I'm not sure about displaying the Block Code editor for if the parent node with BlockCodeNode attached is selected in the Scene Tree. Here I tried the branch adding blocks to AnimationPlayer and TileMap. When I click these nodes, sometimes I get their corresponding editors (Animation bottom dock, TileMap bottom dock) and sometimes the blocks editor.

dylanmccall:
> @dylanmccall this is a huge improvement! I'm not sure about displaying the Block Code editor for if the parent node with BlockCodeNode attached is selected in the Scene Tree. Here I tried the branch adding blocks to AnimationPlayer and TileMap. When I click these nodes, sometimes I get their corresponding editors (Animation bottom dock, TileMap bottom dock) and sometimes the blocks editor.

Hm, okay, I was wondering if that might cause conflicts. I'll change this so the editor only _opens itself_ if the user chooses a Block Code node explicitly, which should solve the problem there and is better in line with how the Animation and Tile Map panels behave.

wnbaum:
I just tested this out on my end. I like the behavior of opening the Block Code panel when a BlockCode node is selected, but just switching the script (not opening the panel) when a parent to a BlockCode node is selected. I think I like this layout better - on my small laptop it is a few extra clicks to get it looking organized, but on larger screens it's way better. It would be good to get some feedback from the learning team to see what they think about the new layout. I wonder if they think being able to see the scene/properties of nodes is worth the layout change.

dylanmccall:
This got a +1 from @jofilizola. I added a couple bits based on her feedback:

- Selecting a BlockCode node causes its parent node to be selected, which gives us better feedback in the 2D canvas.
- (To go with this, I added a way in the Block Code editor to delete a block code script without having to fight with the Scene panel).
- I made that "Add Block Code" button stand out against the background.

Happy to merge this any time :)

wnbaum (approved):
Awesome stuff! Once I get my zooming and panning merged I think it will go really great with this.

One thing I noticed is that when selecting a `BlockCode` node there is a significant amount of lag on my end, presumably from loading all the blocks in (not the fault of this PR at all). We should figure out what's causing that.

PR #119, opened 2024-07-05
Add blockcode warnings

dylanmccall:
This adds some new warnings to `BlockCode` to deal with common problems users may encounter:

Screenshot from 2024-07-05 13-06-08

Note that the "parent should only contain on BlockCode" warning also occurs when the parent is a separate (instantiated) scene. I simulated that in the screenshot by adding a `BlockCode` node to space.tscn, so in pong_game.tscn, where I added another `BlockCode` under `Space`, we get a warning.


wnbaum (approved):
Nice! Good thinking about BlockCode nodes that could be attached in instantiated scenes. Looks good to me.

PR #120, opened 2024-07-05
Add zoom and pan to block canvas

wnbaum:
This adds zooming and panning to the block canvas.

Currently, blocks only change size when they are placed. Is this desired behavior? Or should the blocks shrink/grow as soon as they enter the canvas.

If anyone wants to try this out feedback would be appreciated!

https://phabricator.endlessm.com/T35539


manuq:
This is working like a charm already!

[Grabación de pantalla desde 2024-07-08 12-47-33.webm](https://github.com/endlessm/godot-block-coding/assets/83944/981f2bff-8104-4907-83ec-e414ac38b707)

This also makes the plugin a lot nicer to work in small screens. The infinite canvas is welcome to organize blocks as users wish too.

Is there anything preventing the dragging to be consistent with the 2D viewport? Why the shift modificator instead of middle mouse button?

wnbaum:
>Is there anything preventing the dragging to be consistent with the 2D viewport? 

The `DragManager` is above the node canvas so nodes attached to it aren't scaled with the `Window` node in the canvas. I feel like it would be a nice touch to get the blocks to change scale when you bring them over the canvas though, I'll see if I can get that to work.

>Why the shift modificator instead of middle mouse button?

I was on a trackpad when working on this, and shift-dragging worked the best for panning around. So it makes a bit more accessible that way but I can also have it work with middle mouse button.

Will make some commits.

wnbaum:
It was actually way easier to make these changes than I thought it would be! Partly thanks to @dylanmccall's reorganization of the drag manager.

I really like how it looks now and I think it feels super polished!

dylanmccall (approved):

wnbaum:
Just rebased for the bottom panel update. Working the same as it was! Will merge now.

PR #122, opened 2024-07-08
Add version to block script data

manuq:
https://phabricator.endlessm.com/T35547

### DIFF
--- a/addons/block_code/ui/constants.gd
+++ b/addons/block_code/ui/constants.gd
@@ -1,5 +1,7 @@
 extends Object
 
+const CURRENT_DATA_VERSION = 0
+
 const KNOB_X = 10.0
 const KNOB_W = 20.0
 const KNOB_H = 5.0
--- a/addons/block_code/ui/main_panel.gd
+++ b/addons/block_code/ui/main_panel.gd
@@ -15,6 +15,8 @@ extends Control
 @onready var _icon_collapse := EditorInterface.get_editor_theme().get_icon("Back", "EditorIcons")
 @onready var _icon_expand := EditorInterface.get_editor_theme().get_icon("Forward", "EditorIcons")
 
+const Constants = preload("res://addons/block_code/ui/constants.gd")
+
 var _current_block_code_node: BlockCode
 var _block_code_nodes: Array
 var _collapsed: bool = false
@@ -82,6 +84,14 @@ func _on_delete_dialog_confirmed(block_code_node: BlockCode):
 	undo_redo.commit_action()
 
 
+func _try_migration():
+	var version: int = _current_block_code_node.block_script.version
+	if version == Constants.CURRENT_DATA_VERSION:
+		# No migration needed.
+		return
+	push_warning("Migration not implemented from %d to %d" % [version, Constants.CURRENT_DATA_VERSION])
+
+
 func switch_scene(scene_root: Node):
 	_title_bar.scene_selected(scene_root)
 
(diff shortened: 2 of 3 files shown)


dbnicholson:
I very much like this idea, but I think we may not want to use the actual plugin version in the script. Instead, I think we should maintain a separate data version. Since the data format won't change with every plugin release, we don't need to attempt migration. This would also handle a development snapshot since the data version could be bumped independently of the plugin version.

The data version could be a simple integer counter that we bump whenever the data format changes. We'd set that to 1 now and any script that does not have a version is treated as 0. I don't think we'd need point releases as the semantics would always be to migrate the data to the current version regardless of what it previously was..

manuq:
@dbnicholson good points. I addressed the comments and force-pushed.

addons/block_code/ui/main_panel.gd
```diff
@@ -82,6 +84,14 @@ func _on_delete_dialog_confirmed(block_code_node: BlockCode):
 	undo_redo.commit_action()
 
 
+func _try_migration():
+	var version: int = _current_block_code_node.block_script.version
+	if version == Constants.CURRENT_DATA_VERSION:
+		# No migration needed.
+		return
+	push_warning("Migration not implemented from %d to %d" % [version, Constants.CURRENT_DATA_VERSION])
```
dbnicholson, line 92: I think this is good enough for now, but in the future we might want to beef this up. If the stored data version is less than the current data version, migrate it. If the stored data version is greater than the current data version, then someone has already migrated the data to a newer plugin version and is now trying to run an older version of the plugin. Since the older plugin version by definition doesn't know about the newer data format, I think the only thing to do there would be to refuse to load the script and bring up some kind of error message.

addons/block_code/ui/main_panel.gd
```diff
@@ -129,6 +141,7 @@ func save_script():
 	var generated_script = _block_canvas.generate_script_from_current_window(block_script.script_inherits)
 	block_script.block_trees = block_trees
 	block_script.generated_script = generated_script
+	block_script.version = Constants.CURRENT_DATA_VERSION
```
dbnicholson, line 144: Can we arbitrarily update the version in the saved script if it hasn't been migrated? This is probably fine for now but eventually I'd expect `save_script` to bail if `_current_block_code_node.version != Constants.CURRENT_DATA_VERSION`. Maybe `generate_script_from_current_window` is the one doing the heavy lifting there and should return `null` if it can't understand the blocks on screen.

addons/block_code/ui/constants.gd
```diff
@@ -1,5 +1,7 @@
 extends Object
 
+const CURRENT_DATA_VERSION = 1
```
dbnicholson, line 3: Backtracking on my earlier thought, I wonder if this should still be `0` since we haven't changed the serialization yet. #107, for example, would bump this up. Or we just keep it `0` until we're ready to handle stability.

dbnicholson (approved):
This is a good start. I think it might end up changing form when we actually try to tackle migration, but this gets the basic data in place.

addons/block_code/ui/constants.gd
manuq: Yes, let's keep it zero until we are ready to handle migrations.

dbnicholson (approved):
Looks good now. Merging.

addons/block_code/ui/main_panel.gd
manuq: Yes, this is TBD. I guess we'll try a migration and if that fails the blocks shouldn't be editable at all, can't be saved.

PR #123, opened 2024-07-08
Make SimpleCharacter more flexible

manuq:
To allow different movement kinds for different game possibilities. The previous is now the default and named "as top-down". In addition, now "as platformer" and "as spaceship" are available.

The speed is now exposed as a property to prevent the block getting too long with 3 parameters. The speed vector is used differently in each situation:
- for platformer, Y is the jump speed
- for spaceship, X is the rotation speed

For platformer, the gravity is read from project settings, but also there is a method to update it.

Remove the gravity custom setting from the project setting, that was added back at some point by mistake.

https://phabricator.endlessm.com/T35549

### DIFF
--- a/addons/block_code/simple_nodes/simple_character/simple_character.gd
+++ b/addons/block_code/simple_nodes/simple_character/simple_character.gd
@@ -5,6 +5,9 @@ extends CharacterBody2D
 @export var texture: Texture2D:
 	set = _set_texture
 
+@export var speed: Vector2 = Vector2(300, 300):
+	set = _set_speed
+
 const PLAYER_KEYS = {
 	"player_1":
 	{
@@ -22,6 +25,11 @@ const PLAYER_KEYS = {
 	}
 }
 
+# Get the gravity from the project settings to be synced with RigidBody nodes.
+var gravity = ProjectSettings.get_setting("physics/2d/default_gravity")
+
+var _jumping = false
+
 
 func _set_texture(new_texture):
 	texture = new_texture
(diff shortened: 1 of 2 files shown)


manuq:
Here is a demo:
[Grabación de pantalla desde 2024-07-08 12-31-57.webm](https://github.com/endlessm/godot-block-coding/assets/83944/537898c8-0fda-4a97-8f6e-b751c13dd68d)

manuq:
[Grabación de pantalla desde 2024-07-08 12-31-57.webm](https://github.com/endlessm/godot-block-coding/assets/83944/5ced1d11-af15-49e5-9d2f-220c96bc13d2)

starnight:
The GDScript part is LGTM.  However, should the Pong example's Paddle use the new block `Move with {player: OPTION} buttons as {kind: OPTION}`, too?

wnbaum (approved):
Looks good to me, working as expected.

One comment I have is that couldn't these blocks just be dropdowns in the inspector? Like a `player` dropdown and a `movement type` dropdown as an exported property. The only thing we are really reinforcing with the block scripts is that "movement goes in the physics_process".

I think we really need to think about our target audience, and determine what we want the user to learn from their experience with block coding. I talked to Justin a bit today, and he had similar thoughts - while we want to make the experience fun, we do actually want to teach something about programming. To me this just feels like hiding it away. Things like "set the velocity" and then "moving and sliding" should make sense to high schoolers if explained properly.

What we have here is consistent with the other blocks though, so it seems useful to merge.

manuq:
@starnight thanks, I prevented changing the example for now.

@wnbaum exposing the 3 modes in the Inspector is interesting, we can try that after merge. As per level of abstraction, there are some discussions happening. Justin can formalize that talk you both just had and that will be very appreciated. My understanding from past conversations is that we don't want that lower level "change velocity + move and slide" blocks.

manuq:
I'm going to merge this as the Inspector change can happen after some learning feedback.

PR #127, opened 2024-07-09
T35541 cleanups

dylanmccall:
https://phabricator.endlessm.com/T35541


manuq (approved):
I was testing main today and noticed that now selecting the BlockCode node also always selects the parent / affected node. I can understand the rationale behind it, but I also found it a bit controversial and different to what other plugins do. In any case, this PR lets me select the BlockCode node again by clicking twice on it in the Scene Tree. Allowing me to inspect the underlaying block script data resource in the Inspector again for debugging. So +1 from my side with this cleanup. All changes look fine.

PR #129, opened 2024-07-12
Set MOUSE_FILTER_IGNORE for blocks where necessary

dylanmccall:
Certain components inside our Block scenes were consuming input events, making it difficult to interact with other blocks positioned around them.

### DIFF
--- a/addons/block_code/ui/blocks/statement_block/statement_block.tscn
+++ b/addons/block_code/ui/blocks/statement_block/statement_block.tscn
@@ -7,20 +7,22 @@
 
 [node name="StatementBlock" type="MarginContainer"]
 size_flags_horizontal = 0
+mouse_filter = 2
 script = ExtResource("1_6wvlf")
-defaults = null
 block_name = "statement_block"
 label = "StatementBlock"
 bottom_snap_path = NodePath("VBoxContainer/SnapPoint")
 
 [node name="VBoxContainer" type="VBoxContainer" parent="."]
 layout_mode = 2
+mouse_filter = 2
 theme_override_constants/separation = 0
 
 [node name="TopMarginContainer" type="MarginContainer" parent="VBoxContainer"]
 custom_minimum_size = Vector2(0, 30)
 layout_mode = 2
 size_flags_horizontal = 0
+mouse_filter = 2
 theme_override_constants/margin_left = 0
 theme_override_constants/margin_top = 0
 theme_override_constants/margin_right = 0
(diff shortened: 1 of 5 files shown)


addons/block_code/ui/blocks/utilities/snap_point/snap_point.tscn
```diff
@@ -6,10 +6,9 @@
 anchors_preset = 15
 anchor_right = 1.0
 anchor_bottom = 1.0
-offset_right = -1152.0
-offset_bottom = -648.0
```
dbnicholson, line 10: This seems unrelated. Is it intentional?

addons/block_code/ui/blocks/statement_block/statement_block.tscn
```diff
@@ -32,7 +34,6 @@ layout_mode = 2
 mouse_filter = 1
 script = ExtResource("2_lctqt")
 color = Color(1, 1, 1, 1)
-outline_color = Color(0.8, 0.8, 0.8, 1)
```
dbnicholson, line 35: Ack. I should have resaved this scene when removing the `outline_color` property in cc307af93c21f39bf695f205f8731385e703708a.

addons/block_code/ui/blocks/statement_block/statement_block.tscn
```diff
@@ -54,6 +55,5 @@ theme_override_constants/separation = 0
 
 [node name="SnapPoint" parent="VBoxContainer" instance=ExtResource("3_5vaov")]
 layout_mode = 2
-block_path = NodePath("../..")
```
dbnicholson, line 57: Ack. The `block_path` removal is from f91ac7ee93fadda3fc26ca2f3f522aae7a008c53.

dbnicholson (commented):
I trust this is a good thing to do, but I don't understand why the SnapPoint offset was removed.

addons/block_code/ui/blocks/utilities/snap_point/snap_point.tscn
dylanmccall: Yeah, not intentional. Not sure what that's about - it doesn't really look _right_ - but no reason to change it here. I removed that change now :)

dbnicholson: Maybe it's something that was set in the scene at some point but actually had no effect? 🤷

dbnicholson (approved):

PR #132, opened 2024-07-15
Test serialization

dbnicholson:
This is an effort to start testing how the serialization works. In actuality, I only tested how the GDScript is generated and didn't get to how the resource is serialized. I hope to look at that next but need to understand those parts better first. In order to run this from headless tests, some refactoring was needed. In a nutshell, I want to move as much of the data handling out of the UI elements with proper interfaces.

This will conflict with #107. I'm happy to hold off on this until that lands.


tests/test_instruction_tree.gd
```diff
@@ -63,5 +79,140 @@ func test_tree_node_text():
...
+
+
+func test_basic_script():
+	var ready_block: Block = dup_node(general_blocks["ready_block"])
+
+	var print_block: Block = dup_node(general_blocks["print"])
+	# XXX: It seems like this should substitute {text} in the statement,
+	# but it doesn't. I can't make sense of StatementBlock.
+	# print_block.param_input_strings = {"text": "this is a test"}
+	# print_block._ready()
```
dylanmccall, line 128: I puzzled over this for a little bit, and this error seems pertinent:

```
SCRIPT ERROR: Invalid set index 'text' (on base: 'Nil') with value of type 'String'.
          at: ParameterInput.set_raw_input (res://addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd:57)
```

`GutTest` is a `Node`, so we can call `add_child(ready_block)` instead of having to call `print_block._ready()` and I think that moves things a little closer. I also changed this…

```
	ready_block.bottom_snap.insert_snapped_block(print_block)
	ready_block.bottom_snap.snapped_block = print_block
```

to…

```
	ready_block.bottom_snap.add_child(print_block)
```

… but I still ran into an issue where `SnapPoint._on_child_entered_tree` isn't being called, even though SnapPoint is definitely initialized.

dbnicholson: Yeah, I completely missed the errors when I was running it until today :roll_eyes: I have a refactor in `ParameterInput` that allows it to work without the UI components but it's a bit messy. Since I have to rebase this after #107, I'll try to clean it up and add it.

tests/test_instruction_tree.gd
```diff
@@ -63,5 +79,140 @@ func test_tree_node_text():
...
+
+	# XXX: Why does insert_snapped_block add_child but not set snapped_block?
```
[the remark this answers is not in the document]
dbnicholson: Good to know. I got a little lost between `SnapPoint` and `DragManager` on what the exact semantics were there. Among other things, one of the reasons I like to write these tests is it often forces you to tighten up the API. I've seen several places where some object gets mutated in a seemingly unrelated place. That's completely understandable for where this project is at, but I'd like to clean up some of those parts.

addons/block_code/ui/blocks/block/block.gd
```diff
@@ -21,16 +21,13 @@ signal modified
 @export var category: String
 
 ## The next block in the line of execution (can be null if end)
-@export var bottom_snap_path: NodePath
+@export var bottom_snap: SnapPoint
```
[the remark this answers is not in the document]
dbnicholson: I wouldn't have even noticed or bothered except it was making it a lot harder to test it without manually calling `_ready()`.

addons/block_code/block_code_plugin.gd
```diff
@@ -24,7 +24,6 @@ const DISABLED_CLASSES := [
 	"StatementBlock",
 	"DragDropArea",
 	"SnapPoint",
-	"NodeBlockCanvas",
```
manuq, line 27: :+1: not worth the generalization, there is a single kind of canvas.

addons/block_code/instruction_tree/instruction_tree.gd
```diff
@@ -14,13 +14,13 @@ class TreeNode:
 		children.append(node)
 
 
-func generate_text(root_node: TreeNode, start_depth: int = 0) -> String:
+static func generate_text(root_node: TreeNode, start_depth: int = 0) -> String:
```
manuq, line 17: :+1: yes, these should be static.

addons/block_code/ui/blocks/block/block.gd
```diff
@@ -21,16 +21,13 @@ signal modified
...
 
 ## The scope of the block (statement of matching entry block)
 @export var scope: String = ""
 
-var bottom_snap: SnapPoint
-
 
 func _ready():
-	bottom_snap = get_node_or_null(bottom_snap_path)
```
manuq, line 33: :+1: yes! no need to do this.

manuq (approved):
Besides the initial tests, this sets a good floor for improving the data model. And the cleanup commits are very welcome.

Needs a rebase to fix conflicts.

tests/test_instruction_tree.gd
dbnicholson: I did get this working in tests, but I was unsure whether I was breaking the option handling, so I'm going to punt on that part for the moment.

dbnicholson:
Rebased now. Seems to be working...

tests/test_instruction_tree.gd
wnbaum:
```gdscript
print_block.param_input_strings = {"text": "this is a test"}
add_child(print_block) # Add print_block to scene tree to have it initialize properly
remove_child(print_block) # Make sure it's available to snap to the ready_block

ready_block.bottom_snap.insert_snapped_block(print_block) # Attach to ready_block
ready_block.bottom_snap._update_snapped_block_from_children() # Set snapped_block in ready_block
assert_true(ready_block.bottom_snap.has_snapped_block())
assert_eq(ready_block.bottom_snap.get_snapped_block(), print_block)
```

Changing the test to this works, but it's pretty weird, see the explanation of why it works in the comments. What Dylan said is exactly why the error is happening, but this is a solution. Once the new serialization is in, this test will look more clean.

wnbaum (approved):
Tests and changes look good to me. Once we do our serialization differently the `test_X_script` tests will be a lot cleaner, so not to worry.

tests/test_instruction_tree.gd
dbnicholson: Oh, duh. I had convinced myself that `_ready` would never be called in these tests even though I knew I'd seen it run before in these contexts. Turns out GUT [has a helper](https://gut.readthedocs.io/en/latest/Memory-Management.html#using-add-child-in-tests) to add the child and have it automatically removed and freed at the end of the test.

On the other hand, I was make `ParameterInput` work outside of a scene by storing `raw_input` in the object and reworking when it's transferred to and from the UI controls. I'll try to polish that up, but if I can't I'll do it this way.

PR #134, opened 2024-07-16
Add _to_string for Block

dylanmccall:
The to_string implementation for Block follows the same pattern as the default for Godot objects, but it uses get_block_class and includes the block_name property for each block. This will make it easier to differentiate between different blocks while debugging the block code editor.

-----

This is inspired by #132, where we add block_name for all of the blocks.


manuq:
Yes! I think this good to merge now, before #132 as all blocks have a generic block_name like "statement_block".

PR #135, opened 2024-07-16
Add rich-text tooltips for blocks

wjt:
This overrides the tooltip of all blocks to be a custom control (a thin wrapper around RichTextLabel) and adds a couple of new tooltips that take advantage of this:

image

I don't want to spend too much time on the tooltips themselves because we intend to review the wording of the blocks themselves & terminology used in conjuction with the learning team – and potentially changing the level of abstraction of some of the blocks. Similarly I could imagine making the tooltips fancier by, for example, listing the parameters, adding a note about what kind of block it is, etc. But again I think this would be better done in conjuction with the learning team. So here I'm focusing on just having the capability to use rich text in tooltips.

https://phabricator.endlessm.com/T35540

### DIFF
--- a/addons/block_code/ui/blocks/block/block.gd
+++ b/addons/block_code/ui/blocks/block/block.gd
@@ -74,3 +74,9 @@ func serialize_props(prop_names: Array) -> Array:
 	for p in prop_names:
 		pairs.append([p, self.get(p)])
 	return pairs
+
+
+func _make_custom_tooltip(for_text) -> Control:
+	var tooltip = preload("res://addons/block_code/ui/tooltip/tooltip.tscn").instantiate()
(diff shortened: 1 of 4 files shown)


manuq:
I spent some time trying to find a better option than doing the font overrides in a script, but it seems this is the way to go.

Changing the editor main font worked too:
Captura desde 2024-07-16 08-48-19

wjt:
> I spent some time trying to find a better option than doing the font overrides in a script,

I know, right? I spent a long time trying to find a better way.

In the end I searched GitHub for `@tool` and `RichTextLabel` and this really does seem to be the most common way.

Maybe I'll turn to Discord.

PR #137, opened 2024-07-16
Serialization: Reuse SerializedBlock resources

manuq:
Add a resource property to blocks of type SerializedBlock. The resource will be updated before building the tree.

This mitigates having a huge diff in scenes with blocks for minimum changes like moving a block in the canvas.

https://phabricator.endlessm.com/T35565

### DIFF
--- a/addons/block_code/ui/block_canvas/block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/block_canvas.gd
@@ -100,8 +100,9 @@ func bsd_selected(bsd: BlockScriptData):
 
 	var edited_node = EditorInterface.get_inspector().get_edited_object() as Node
 
-	_window.position = Vector2(0, 0)
-	zoom = 1
+	if bsd != _current_bsd:
+		_window.position = Vector2(0, 0)
+		zoom = 1
 
 	_window.visible = false
 	_zoom_label.visible = false
--- a/addons/block_code/ui/block_canvas/block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/block_canvas.gd
@@ -40,6 +40,8 @@ var zoom: float:
 	get:
 		return _window.scale.x
 
+var _undo_redo: EditorUndoRedoManager
+
 signal reconnect_block(block: Block)
 signal add_block_code
 signal open_scene
@@ -170,24 +172,29 @@ func load_tree(parent: Node, node: SerializedBlockTreeNode):
 		load_tree(scene.get_node(c[0]), c[1])
 
 
-func get_canvas_block_trees() -> SerializedBlockTreeNodeArray:
-	var block_trees := SerializedBlockTreeNodeArray.new()
+func rebuild_block_trees(undo_redo):
+	_undo_redo = undo_redo
+	_current_bsd.block_trees.array = []
 	for c in _window.get_children():
-		block_trees.array.append(build_tree(c))
-
-	return block_trees
+		_current_bsd.block_trees.array.append(build_tree(c))
 
 
 func build_tree(block: Block) -> SerializedBlockTreeNode:
-	var n = SerializedBlockTreeNode.new()
-	n.serialized_block = SerializedBlock.new(block.get_block_class(), block.get_serialized_props())
+	var path_child_pairs = []
+	block.update_resources(_undo_redo)
 
 	for snap in find_snaps(block):
-		for c in snap.get_children():
-			if c is Block:  # Make sure to not include preview
-				n.path_child_pairs.append([block.get_path_to(snap), build_tree(c)])
+		for child in snap.get_children():
+			if not child is Block:  # Make sure to not include preview
+				continue
+			path_child_pairs.append([block.get_path_to(snap), build_tree(child)])
 
-	return n
+	if block.resource.path_child_pairs != path_child_pairs:
+		_undo_redo.add_undo_property(block.resource, "path_child_pairs", block.resource.path_child_pairs)
+		block.resource.path_child_pairs = path_child_pairs
+		_undo_redo.add_do_property(block.resource, "path_child_pairs", block.resource.path_child_pairs)
+
+	return block.resource
 
 
 func find_snaps(node: Node) -> Array:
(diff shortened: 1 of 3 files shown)


manuq:
We can do something similar to SerializedBlockTreeNode and SerializedBlockTreeNodeArray. But this totally breaks the undo/redo. The benefit of having the tree of resources being rebuilt each time is that it makes undo/redo so trivial. So marking this as draft for now.

dbnicholson:
I certainly like that. Can you pass the `EditorUndoRedoManager` down through `get_canvas_block_trees`?

manuq:
> I certainly like that. Can you pass the `EditorUndoRedoManager` down through `get_canvas_block_trees`?

Good point, yes I did so. Now the undo/redo works with block properties, like changing a parameter or moving the block to another position.

Then I tried to do the same for the children blocks (path_child_pairs), but is now behaving correctly. I'm running out of ideas, it seems to be the tree being built recursively. Any help welcome!

addons/block_code/ui/main_panel.gd
```diff
@@ -139,17 +139,14 @@ func save_script():
...
-
-	var block_trees := _block_canvas.get_canvas_block_trees()
+	_block_canvas.rebuild_block_trees(undo_redo)
 	var generated_script = _block_canvas.generate_script_from_current_window(block_script)
-	block_script.block_trees = block_trees
-	block_script.generated_script = generated_script
+	if generated_script != block_script.generated_script:
+		undo_redo.add_undo_property(_current_block_code_node.block_script, "generated_script", _current_block_code_node.block_script.generated_script)
+		block_script.generated_script = generated_script
+		undo_redo.add_do_property(_current_block_code_node.block_script, "generated_script", _current_block_code_node.block_script.generated_script)
```
dylanmccall, line 147: We shouldn't be changing the generated_script property on our own here, and we should consistently use the `var block_script: BlockScriptData` we defined earlier:

```suggestion
	if generated_script != block_script.generated_script:
		undo_redo.add_undo_property(block_script, "generated_script", block_script.generated_script)
		undo_redo.add_do_property(block_script, "generated_script", generated_script)
```

dylanmccall: Turns out we're doing this in a lot of places. I included this and a bunch of similar fixes in #140.

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -38,6 +38,8 @@ var zoom: float:
 	get:
 		return _window.scale.x
 
+var _undo_redo: EditorUndoRedoManager
```
dbnicholson, line 41: Personally I would just pass this down in to `build_tree` rather than stashing it in the object. It's only called internally to this object, so there's no danger of it breaking another user.

manuq: Yes I though about that, I did it like this because is also used in the `build_tree()` method that's called recursively. But yes it can be passed around to `build_tree()` as parameter.

manuq:
This is now ready for review thanks to @dylanmccall fixes. The undo-redo is now piling actions to the history instead of going back as before. I couldn't find how to fix it. But is usable (can undo, can redo).

manuq:
> This is now ready for review thanks to @dylanmccall fixes. The undo-redo is now piling actions to the history instead of going back as before. I couldn't find how to fix it. But is usable (can undo, can redo).

The undo-redo piling actions was an issue in the input parameter, fixed in: https://github.com/endlessm/godot-block-coding/pull/142

addons/block_code/ui/blocks/block/block.gd
```diff
@@ -61,6 +64,17 @@ func get_instruction_node() -> InstructionTree.TreeNode:
 	return node
 
 
+func update_resources(undo_redo: EditorUndoRedoManager):
+	if resource == null:
+		resource = SerializedBlockTreeNode.new()
+		resource.serialized_block = SerializedBlock.new(get_block_class(), get_serialized_props())
```
dbnicholson, line 70: I think you can return early here. There's no need to compare the serialized properties again if the serialized block was just constructed with them.

addons/block_code/ui/blocks/block/block.gd
```diff
@@ -61,6 +64,17 @@ func get_instruction_node() -> InstructionTree.TreeNode:
...
+
+	var serialized_props = get_serialized_props()
+	if serialized_props != resource.serialized_block.serialized_props:
+		undo_redo.add_undo_property(resource.serialized_block, "serialized_props", resource.serialized_block.serialized_props)
+		undo_redo.add_do_property(resource.serialized_block, "serialized_props", serialized_props)
```
dbnicholson, line 75: One thing I thought about this morning with an eye towards the future decoupled data and UI is that maybe we don't want to modify the resources in place. If we want to get to a place where the UI only refers to a code resource, then I think that's easier to handle if the serialized block is treated immutably.

In other words, rather than updating `serialized_props` on an existing `SerializedBlock`, detect if `serialized_props` has changed and replace the `serialized_block` property in that case. Like:

```
if resource == null:
    resource = SerializedBlockTreeNode.new()

if resource.serialized_block == null:
    resource.serialized_block = SerializedBlock.new(get_block_class(), get_serialized_props())
    return

var serialized_props = get_serialized_props()
if serialized_props != resource.serialized_block.serialized_props:
    undo_redo.add_undo_property(resource, "serialized_block", resource.serialized_block)
    undo_redo.add_do_property(resource, "serialized_block", SerializedBlock.new(get_block_class(), get_serialized_props()))
```

WDYT?

manuq: @dbnicholson hmm, this will always generate a new resource ID for SerializedBlock, which is the opposite this PR is trying to do. What would be exactly the benefit?

dbnicholson: It's really in between both worlds. If the block inputs haven't changed, leave it as is. If the have, replace the object. The idea being that if there are multiple objects referencing this serialized block, then they may not be prepared to have it modified in place. That's just theoretical, though, as the serialized blocks are always referenced from a single UI block right now. So, let's just move on.

dbnicholson (approved):
Looks good to me. From my perspective, feel free to merge after squashing the fixup.

PR #138, opened 2024-07-16
tooltip: Don't save style overrides to tooltip.tscn

wjt:
Previously, if you opened tooltip.tscn in the editor and hit Save, its `_ready()` method would persist the style overrides we use to apply the editor theme's fonts to the label. This isn't necessary, or desirable.

Godot 4.3+ exposes `Node.is_part_of_edited_scene()` to GDScript, which is analogous to `Engine.is_editor_hint()` but determines whether the `@tool` script's node is itself being edited, as opposed to instantiated elsewhere in the editor.

Reimplement this function in GDScript and only apply the style overrides to the tooltip if it returns false.

https://phabricator.endlessm.com/T35540


addons/block_code/ui/tooltip/tooltip.gd
```diff
@@ -22,3 +22,13 @@ func _ready():
 	bold_italics.set_base_font(italics)
 	bold_italics.set_variation_embolden(1.2)
 	add_theme_font_override(&"bold_italics_font", bold_italics)
+
+
+## Polyfill of Node.is_part_of_edited_scene(), available to GDScript in Godot 4.3+.
+static func node_is_part_of_edited_scene(node: Node) -> bool:
+	return Engine.is_editor_hint() && node.is_inside_tree() && node.get_tree().edited_scene_root && node.get_tree().edited_scene_root.get_parent().is_ancestor_of(node)
```
wjt, line 29: When I wrap this incredibly long line, `gdformat` insists on changing it back...

wjt:
I pushed a couple more changes to apply the same fix to the scenes discussed on Slack.

addons/block_code/ui/util.gd
```diff
@@ -0,0 +1,6 @@
+extends Node
```
dbnicholson, line 1: I think you should extend `Object` if you're just going to put static functions in here. Also, I think you might as well add this first in the series and then use it in tooltip.gd.

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -45,7 +47,7 @@ signal replace_block_code
 
 
 func _ready():
-	if not _open_scene_button.icon:
+	if not Util.node_is_part_of_edited_scene(self) and not _open_scene_button.icon:
```
dbnicholson, line 50: Nitpick - optimize the logic by moving the simple `null` check first.

addons/block_code/ui/tooltip/tooltip.gd
dbnicholson: Even with parentheses?

dbnicholson (changes requested):
Wow, nice find. I have often wondered why so much stuff is getting serialized into the scenes.

addons/block_code/ui/util.gd
wjt: OK

addons/block_code/ui/block_canvas/block_canvas.gd
wjt: I think the null check is now redundant: it will always be null when _ready runs.

addons/block_code/ui/tooltip/tooltip.gd
wjt: Yes, amazingly enough

wnbaum (commented):
Looks good to me once you fix Dan's changes! This is a much more elegant solution than restoring the file every time :)

addons/block_code/ui/tooltip/tooltip.gd
wjt: Oh right, it's because the maximum line length was set to 200 chars in e29f171f.

addons/block_code/ui/block_canvas/block_canvas.gd
wjt: I left the null check in, but swapped the condition around.

dbnicholson (approved):
Sweet

manuq:
I'm late to the party, this is a fantastic improvement!

addons/block_code/ui/blocks/parameter_block/parameter_block.tscn
```diff
@@ -1,19 +1,8 @@
-[gd_scene load_steps=4 format=3 uid="uid://clipm2dd28jde"]
+[gd_scene load_steps=3 format=3 uid="uid://clipm2dd28jde"]
 
 [ext_resource type="Script" path="res://addons/block_code/ui/blocks/parameter_block/parameter_block.gd" id="1_0hajy"]
 [ext_resource type="PackedScene" uid="uid://c7puyxpqcq6xo" path="res://addons/block_code/ui/blocks/utilities/drag_drop_area/drag_drop_area.tscn" id="2_gy5co"]
 
-[sub_resource type="StyleBoxFlat" id="StyleBoxFlat_dbera"]
-bg_color = Color(1, 1, 1, 1)
```
wnbaum, line 7: Wait were these StyleBoxes removed manually? They are meant to be saved to the scene, somehow I missed this.

addons/block_code/ui/picker/categories/block_category_button.tscn
```diff
@@ -1,14 +1,7 @@
-[gd_scene load_steps=7 format=3 uid="uid://bdtetj0gs45hv"]
+[gd_scene load_steps=6 format=3 uid="uid://bdtetj0gs45hv"]
 
 [ext_resource type="Script" path="res://addons/block_code/ui/picker/categories/block_category_button.gd" id="1_pxxnl"]
 
-[sub_resource type="StyleBoxFlat" id="StyleBoxFlat_eogpc"]
-bg_color = Color(1, 0, 0, 1)
```
wnbaum, line 6: Here too.

addons/block_code/ui/blocks/parameter_block/parameter_block.tscn
wjt: Hmm I think I reset them to defaults in the editor on the basis that they will be instantiated every time the scene is actually used. I didn't see the square-icon issue... I can investigate tomorrow

wnbaum: Yeah, the default stylebox value is actually used as a template, only the color is set on it. Which is kind of weird, maybe we should just create a totally new stylebox at runtime.

PR #142, opened 2024-07-17
Parameter input: change text modification logic

manuq:
Previously any text change was sending the modified signal, adding too many actions to the undo/redo history (one per typed character). This was also capturing shortcuts like Ctrl-Z for undoing and sending the modified signal, thus piling to the history, not going back.

Instead, send the modified signal when the line edit is submitted (enter key pressed) or when it loses focus and the text has changed.

https://phabricator.endlessm.com/T35566

### DIFF
--- a/addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
+++ b/addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
@@ -31,6 +31,11 @@ var option: bool = false
 @onready var _bool_input := %BoolInput
 @onready var _bool_input_option := %BoolInputOption
 
+# Used to submit the text when losing focus:
+var _last_lineedit_submitted_text: String
+var _last_x_lineedit_submitted_text: String
+var _last_y_lineedit_submitted_text: String
+
 
 func set_raw_input(raw_input):
 	if option:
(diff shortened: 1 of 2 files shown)


addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
```diff
@@ -126,10 +131,16 @@ func get_string() -> String:
 			return "%s" % input
 
 
-func _on_line_edit_text_changed(new_text):
+func _on_line_edit_text_submitted(new_text):
```
dbnicholson, line 134: This function is also used to handle the Vector2 input signals. You'll need to split those to a separate function (arguably they should have been already).

dbnicholson (changes requested):

addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
manuq: Oh I totally missed that, good catch! PR updated. It is a bit boilerplate now. An alternative would be to connect to the callbacks in code (not in the Signals dock in the editor) and bind the corresponding LineEdit as argument.

dylanmccall (approved):
This is working well for me now.

manuq:
@dbnicholson do you think this is good for merging now?

dbnicholson (approved):
Sorry, forgot to check back on this! Yes, it looks good now!

PR #144, opened 2024-07-18
Reinstate stylebox overrides

wjt:
This partially reverts commit 1b6d64b4f6d532139e29acb0d1cf9c4824ac9ced (“Don't recreate resources when editing plugin scenes”).

I incorrectly believed that these were being fully recreated every time the scene is instantiated based on another element's stylebox, but in fact they are being used as a template and it is only the colour that is being adjusted.

See discussion at https://github.com/endlessm/godot-block-coding/pull/138#discussion_r1681557942.


wnbaum (approved):
Thanks for fixing! Looks good to me.

PR #147, opened 2024-07-20
Decouple UI from Code Generation

wnbaum:
Decouples UI from code generation. On top of #190.

So far:
1. Load block resources from `.tres` files
2. Replace category factory block creation with simple block resources with same properties
3. Load blocks by name from block script data. Make them into an AST. Then create the UI from the AST.
4. Create updated AST from UI when modified. Uses resources stored in UI.
5. Save blocks by name from AST into block script data.
6. Serialize user input (defaults and user input actually serialized as Variants)
7. Merge ASTs if same entry statement
8. Fix option inputs
9. Fix variable blocks being loaded from name (need to load custom resources from BSD before loading script)
10. Add old blocks back
11. Remake pong game
12. Fix scope
13. Fix block highlighting
14. Fix undo

To do:
1. ~Replace my temporary block resource with Manuel's (for overloading). Make sure things like control blocks with multiple child snaps work~. (Lets save this for another PR where we can modify the block resource for overloading).
2. ~Reimplement unique local variables~ (Can also be another PR)
3. Clean up

https://phabricator.endlessm.com/T35591


wnbaum:
Did some more work on this. If someone does want to go through the code you can but it's still pretty messy. You can now make block scripts again, and they save and load just fine. The old blocks are not ported yet though.

The idea of this is to have one central data structure, `BlockAST` that generates the code and stores the structure of block resources. The UI is generated from the AST, and the serialized scripts get saved from the AST. The AST is loaded from the serialized script as well. The serialized script is in the form of a "BlockNameTree" which is what it sounds like, a tree of named blocks (along with raw user input arguments) which should be the minimum amount of data we need to save for a block script.

So a little diagram would look like this:
```
UI <-> BlockAST <-> BlockNameTree
          |
          V
    Generated Code
```

wnbaum:
Um, wow. I did something weird with git xD

dbnicholson:
I want to make sure I understand the data model correctly. It seems like `BlockNameTreeNode` is a minimized version of `BlockAST.ASTNode`. The point of `BlockNameTreeNode` is to not serialize individual `BlockResource`s (which are children of `ASTNode`) to the scene. In other words, `ASTNode` is a concrete instance of `BlockNameTreeNode` + `BlockResource`. Is that correct?

The relationship between `BlockAST` and `BlockScriptData` is a bit confusing. It feels like `BlockScriptData` is the AST root node, but I guess it's more like a wrapper around the various top level ASTs. Is that correct?

Am I wrong or is `InstructionTree` unused after this?

I have to admit that the names are slowing me down. Whenever I see some object that begins with `Block`, I have to dig around to figure out if it's a UI element, serialization element or code element.

Anyways, now that I've dug through most of this, I mostly like it. You can mostly see how the GDScript can be built from objects independently of the UI. I like that the generated script clearly comes out of the AST state rather than the UI state or the serialized state. The downside is that `BlockCanvas` is doing a lot of heavy lifting transitioning between the 3 states.

dbnicholson:
I see that all the `SerializedBlock*` classes are gone, too. One thing that I think would have made this easier to review is if you either used the existing classes or renamed them and then made changes.

wnbaum:
> I want to make sure I understand the data model correctly. It seems like BlockNameTreeNode is a minimized version of BlockAST.ASTNode. The point of BlockNameTreeNode is to not serialize individual BlockResources (which are children of ASTNode) to the scene. In other words, ASTNode is a concrete instance of BlockNameTreeNode + BlockResource. Is that correct?

`ASTNode` stores a `BlockResource` in its data field, and its children are other `ASTNode`s. `BlockNameTreeNode` is also a tree node class, but stores the name of the block as its data, and its children are other `BlockNameTreeNode`s. These `BlockNameTreeNode`s are what get saved directly to the `BlockScriptData`. When a `BlockScriptData` is loaded, an `ASTNode` is created for each corresponding `BlockNameTreeNode`. These loaded `ASTNode`s will now contain the actual `BlockResource` that the `BlockNameTreeNode`'s `block_name` is referencing.

So again, an `ASTNode` is just a tree node that contains the `BlockResource`. The `BlockResource` is what defines the actual code that gets generated in the `BlockAST` later. These `ASTNode`s are created when we match the `BlockNameTreeNode.block_name` field to its corresponding `BlockResource`.

> The relationship between BlockAST and BlockScriptData is a bit confusing. It feels like BlockScriptData is the AST root node, but I guess it's more like a wrapper around the various top level ASTs. Is that correct?

Correct. Just like we stored the `SerializedBlockTree`s as a field in `BlockScriptData` before, we now store `BlockNameTrees`. These `BlockNameTrees` are composed of `BlockNameTreeNode`s that store the bare minimum data they need to: the block name and whatever user input is specified. There is a `BlockNameTree` for each entry block, and they each have a position on the canvas.

> Am I wrong or is InstructionTree unused after this?

Correct. I will remove it. The `BlockAST` does similar things for code generation.

> I have to admit that the names are slowing me down. Whenever I see some object that begins with Block, I have to dig around to figure out if it's a UI element, serialization element or code element.

I definitely think some better class names would be helpful. I'll do this before I take this PR out of draft.

> Anyways, now that I've dug through most of this, I mostly like it. You can mostly see how the GDScript can be built from objects independently of the UI. I like that the generated script clearly comes out of the AST state rather than the UI state or the serialized state. The downside is that BlockCanvas is doing a lot of heavy lifting transitioning between the 3 states.

Yes, there's certainly a lot of nice things this accomplishes! I agree that the `BlockCanvas` is doing too much. I can surely move some tree construction methods to their respective tree classes. I do think having these 3 states is nice though because it is more flexible for code changes in the future. Suppose we want to change the `BlockNameTreeNode` serialization to some sort of markup type language. Then we wouldn't even have to touch the UI or code generation, only the interface between the serialization and the `BlockAST`.

wnbaum:
> I see that all the `SerializedBlock*` classes are gone, too. One thing that I think would have made this easier to review is if you either used the existing classes or renamed them and then made changes.

Ah, very true. I didn't think about that being a way that would point you in the right direction when reviewing. Good to know! We should definitely discuss these decoupling changes tomorrow and either this will have been a good experiment or something I can clean up and merge :) Thanks for checking it out @dbnicholson!

fire:
Hi, I am looking for a script data ast, can point to where that is implemented.

wnbaum:
> Hi, I am looking for a script data ast, can point me to where that is implemented.
> 
> I am investigating adding a code generator to take ast to c to tcc to https://github.com/libriscv/godot-sandbox for runtime headless code execution

Hey @fire, this is a very basic AST that is used to convert our code blocks to GDScript, so most likely not what you are looking for, sorry.

wnbaum:
Rebased on top of Manuel's most recent changes! Working without errors - though I still have to port some stuff over to `BlockCatalog`, reimplement the focus highlighting, and just clean up in general. Will split into comprehensible commit messages as well.

addons/block_code/blocks/logic/else_if.tres
```diff
@@ -0,0 +1,15 @@
...
+
+[ext_resource type="Script" path="res://addons/block_code/code_generation/block_definition.gd" id="1_kgjks"]
+
+[resource]
+script = ExtResource("1_kgjks")
+name = &"else_if"
+type = 4
+variant_type = 0
+display_template = "Else if {condition: BOOL}"
+code_template = "elif {condition}:"
```
manuq, line 11: @wnbaum I wonder why you decided to make "else" or "else if" separate blocks.
Captura desde 2024-08-06 07-51-51

This can create scripts with errors:
Captura desde 2024-08-06 08-05-54

Is it because blocks with multiple slots are hard in this new architecture? In that case it is less flexible than what we had, going against the flexibility goal.

IMO the beauty of blocks is that they prevent parse and syntax errors.

manuq:
> Rebased on top of Manuel's most recent changes! Working without errors - though I still have to port some stuff over to `BlockCatalog`, reimplement the focus highlighting, and just clean up in general. Will split into comprehensible commit messages as well.

That was a fast rebase @wnbaum, cool!

I gave it a try and everything works without errors.

I see all the resource IDs are changing when I change a single text:
Captura desde 2024-08-06 08-11-28


Do you think that regression can be fixed?

wnbaum:
> I see all the resource IDs are changing when I change a single number argument:
> 
> Do you think that regression can be fixed?

Yes! The way the serialized tree is reconstructed, it creates all new resources. This logic is a little different from yours I think because now the `BlockDefinition` is being stored in the UI rather than the `BlockSerialization`. That's why the logic didn't transfer over when I rebased. Anyways, you are correct that this is a regression and I will just check if the resource already exists by traversing the old serialized tree while creating the new one, and that should yield similar results.

addons/block_code/blocks/logic/else_if.tres
wnbaum: Yes, the main reason was ease of implementation. We could accomplish this with different `BlockDefinition` types, like `StatementBlockDefinition` and `ControlBlockDefinition` as well as a more complex abstract syntax tree. As you can see in the `BlockDefinition` class we have a `block_type` and a `variant_type` which lets us combine all the current block types into one data model. This is nice for simplicity, but only allows one statement for control blocks (unless we were to generalize the model even more). 

Also, the AST is currently made up of nodes with children. Each entry block has children, and each control block has children, and the children are the sequence of blocks that are indented under the current statement. We would need to expand the AST, BlockNameTree, and UI generation functions to take into account multiple children groups per node. This wasn't an issue before since we just saved the paths to each block, and generated the code through the UI, which IMO was very ugly.

So it's definitely possible to do undo this regression, but it would add a lot more to this PR. Also, there are some benefits to having the control blocks being like this. If you wanted to convert an `if/else` statement to an `if` in the old version, you would have to take all the blocks out, switch out the control block, then put all the blocks back in. Whereas with this you can just chuck the `else` block. The only thing you would have to do to guarantee correct scripts is make sure `else` and `elif` blocks can only be attached to `if` and `elif` blocks.

Anyways, there's definitely arguments for both UI methods, so I thought it would be best to do the simplest implementation and then decide if it is really worth porting back later. IIRC @dbnicholson had similar thoughts when we talked about it.

wnbaum:
Actually @manuq I'm thinking it might be better to wait and have the changes for keeping the resource IDs the same live in a separate PR. There's many ways to do this as I've found out today: 
1. My original idea was to traverse the old block name tree while creating the new block name tree, and reuse resources with same name. (Hard in this version because we need to preserve the data in the old tree to explore it, while overriding its properties to use in the new tree).
2. Override `_get_rid()` with a hash of each block's tree position. This way RIDs are consistent. (Easier to implement but may not reuse some resources if positions are relocated).
3. Use our own custom serialization that stores block names and parameters in a parseable and readable way  (which I think I prefer as it will make diffs much more readable. After all, we are serializing code and code diffs are what we and git are used to).

The root of the problem is that there is no easy way to get Godot to consistently serialize resources in the exact same position and with the exact same IDs (other than actually storing them in the structure of the UI like `main` is doing currently, _not decoupled_).

It really comes down to how we want to serialize the AST. This PR supports flexibility in that we can choose how to generate the UI from the AST, or how we want to serialize it since the three states are "decoupled" from each other. Do you think it's worth it to try to implement one of those three methods I described now? Or just consider it later.

manuq:
> Do you think it's worth it to try to implement one of those three methods I described now? Or just consider it later.

No, please cleanup what you have so we can compare. Honestly I would like to balance the benefits that this code generation provides VS the current one. Is the cost in complexity acceptable? Is there a shorter path to blocks flexibility? Doing any of the 3 ways you mention just to keep the RIDs the same seems too much, and very different to what (for example) the visual shaders do in Godot.

If you can, provide smaller PRs from the description above. Eg. "1. Load block resources from .tres files" could be split into its own PR. There was concensus to do that among @dbnicholson and @dylanmccall.

wnbaum:
Technically ready for review but we should discuss. Should we split into small PRs? Do we want all of these changes? Or is everything good and clear enough to keep it in one.

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -136,482 +171,192 @@ static func _category_cmp(a: BlockCategory, b: BlockCategory) -> bool:
...
+static func get_categories(blocks: Array[BlockDefinition], extra_categories: Array[BlockCategory] = []) -> Array[BlockCategory]:
 	var cat_map: Dictionary = {}
 	var extra_cat_map: Dictionary = {}
 
 	for cat in extra_categories:
 		extra_cat_map[cat.name] = cat
 
 	for block in blocks:
-		var cat: BlockCategory = cat_map.get(block.category)
+		var block_cat_name: String = block.category
```
manuq, line 182: Can you name this `block_category_name`? Or maybe just `name` would work in this context?

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -136,482 +171,192 @@ static func _category_cmp(a: BlockCategory, b: BlockCategory) -> bool:
...
 
 	const FALLBACK_SET_FOR_TYPE = {
-		TYPE_INT: "0",
-		TYPE_FLOAT: "0",
-		TYPE_VECTOR2: "0,0",
-		TYPE_COLOR: "DARK_ORANGE",
+		TYPE_INT: 0,
+		TYPE_FLOAT: 0.0,
+		TYPE_VECTOR2: Vector2(0, 0),
+		TYPE_COLOR: Color.SKY_BLUE,
```
manuq, line 310: You didn't like DARK_ORANGE? :) I though it played nice with Endless colors.

manuq (changes requested):
Here is a review of the first 4 commits. Well done!

In general I think the commits messages can be more descriptive. For example the body of number 4 can say: "Traverse the blocks/ directory containing .tres files with the new block definitions, and add them to the catalog. Remove the previous definitions."

addons/block_code/serialization/block_name_tree_node.gd
```diff
@@ -0,0 +1,13 @@
+extends Resource
+
+const BlockNameTreeNode = preload("res://addons/block_code/serialization/block_name_tree_node.gd")
```
manuq, line 3: I think we agreed on naming this BlockSerialization. This seems to be a replacement for the current file `addons/block_code/serialization/block_serialization.gd`. It would be nice if you can change that file instead of leaving it abandoned and adding a new one.

addons/block_code/serialization/block_name_tree_node.gd
```diff
@@ -0,0 +1,13 @@
...
+
+@export var name: StringName
+@export var children: Array[BlockNameTreeNode]
```
manuq, line 6: If I understand correctly, this is [path_child_pairs](https://github.com/endlessm/godot-block-coding/blob/55acc2d917eb1250c81d92262f6e7d14a4df45b0/addons/block_code/serialization/block_serialization.gd#L6) in the current serialization. And that allowed to map multiple snap points to children blocks. I think that the new format should also support that. This is why an if/else block is not possible anymore. Ideally this should be a Dictonary that maps slot names to children blocks, so no need to have NodePaths in the serialization like currently, eg from the Pong serialization:
```
path_child_pairs = [[NodePath("VBoxContainer/SnapPoint"), SubResource("Resource_ftveg")]]
```
For example:
- An entry block like "On ready" can have a single slot named "body". So the Dictionary would be: `{ "body": [BlockSerialization, BlockSerialization]}`
- A sentence block like "Print" has no slots, so the Dictionary is empty: `{}`
- An if/else block has 2 bodies: `{"if_body": [BlockSerialization], "else_body": [BlockSerialization, BlockSerialization]}`

Do you think this would be too difficult to change at this point?

addons/block_code/serialization/block_name_tree.gd
```diff
@@ -0,0 +1,24 @@
...
+
+const BlockNameTreeNode = preload("res://addons/block_code/serialization/block_name_tree_node.gd")
+
+@export var root: BlockNameTreeNode
+@export var canvas_position: Vector2
+
+
+func _init(p_root: BlockNameTreeNode = null, p_canvas_position: Vector2 = Vector2(0, 0)):
+	root = p_root
+	canvas_position = p_canvas_position
```
manuq, line 11: This is an interesting change compared with the current serialization. If I understand correctly:
- The block script has an array of these trees
- There is one of these trees per set of blocks attached together in the canvas
- At the same time the blocks have other blocks as children
- Position in the canvas is set for the trees, because there is no need to save position of blocks attached to others, only to the root block.

Currently this extra tree is not needed:
- The block script has an array of entry blocks
- The blocks have children
- All blocks have position, but the only position that's taken into account is the one for the root blocks

I imagine a user dragging a block that's attached to an empty place in the canvas. In the new way, more data will be changed because a new tree resource will be created (more diff). In the current way, ignored data is persisted but is immediately available.

Not a big deal though, just trying to understand the change.

manuq (changes requested):
Feedback for commit 6 (serialization).

addons/block_code/ui/picker/categories/category_factory.gd
wnbaum: Idk why I felt threatened by the orange that day... it looks good to me now haha.

addons/block_code/serialization/block_name_tree_node.gd
wnbaum: Correct, I like the idea of having a dictionary with snap point slot names as keys, and a list of blocks as values.

I don't think it would be too difficult to change if you wanted to do it in this PR. You would have to probably make a subclass of `BlockDefinition` like `MultiBlockDefinition` that has an array of `code_template`s and `display_template`s. Then, when traversing the tree to generate the other tree states, you would have to traverse the children in each slot, as well as all of the arguments for the slot headers (which may be a bit trickier).

I still think it would probably be better to just iterate on this later since we are still a little unsure (personally I found the Scratch-type if/else blocks a little annoying sometimes), and we aren't really losing any functionality here, just some safety checking. And there's plenty to look at for safety checking in the future :grin:. I think the main benefit of getting this PR in will be to separate the states. If you want to reimplement control blocks it will be much less hacky than it was before.

addons/block_code/serialization/block_name_tree.gd
wnbaum: Right, the way the serialization state is done here is with resources. I tried to make it so that the resources contain the bare minimum amount of data to recreate the block trees, but I realize that the resources themselves store quite a bit of data intrinsically as well.

Again I think this might be a good place for iteration. The serialization in this PR is very Godot-like and achieved the goals I wanted (like loading blocks from name and storing minimum data), but it is unfortunate that the resources generate large diffs. I think the `VisualShaderNode`s in the engine could provide some better insight as to the proper way to serialize things like this, so maybe that could be an area of research in the future.

addons/block_code/serialization/block_name_tree_node.gd
wnbaum: Okay, just renamed the serialization classes and propagated the changes to commits on top!

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -128,6 +129,40 @@ const BUILTIN_PROPS: Dictionary = {
 	},
 }
 
+static var block_definition_dictionary: Dictionary
```
manuq, line 132: There is already a dictionary of block definitions in https://github.com/endlessm/godot-block-coding/blob/b11b95ec28e99e0e3e0113501701780958307da6/addons/block_code/code_generation/blocks_catalog.gd#L6 . That should be used instead.

dbnicholson:
I rebased this on aa80f0cbd30616268a1cdb86fe3e65152434a3f8. I've pushed the 2 previous versions of this branch as [decouple-ui-from-codegen-v1](https://github.com/endlessm/godot-block-coding/tree/decouple-ui-from-codegen-v1) and [decouple-ui-from-codegen-v2](https://github.com/endlessm/godot-block-coding/tree/decouple-ui-from-codegen-v2). decouple-ui-from-codegen-v2 is the last branch that @wnbaum pushed.

Here's a high level view of the changes:

* Some things got messed up in the rebase commits. Particularly with all the movement away from `CategoryFactory` to `BlocksCatalog`. I added some more commits that need a bit of cleanup to handle that.
* There were 2 caches in `BlockCanvas` that I removed. One of them was a `BlockDefinition` cache that duplicated what was in `BlocksCatalog`. The other was a `BlockCategory` cache that's mostly unneeded. However, I think it did handle a custom block class providing `get_custom_categories`. That appears to be broken in main and I didn't try to fix it here. I think that needs the same treatment as in 768de50eeb9fbd6899228a8093a9a0ab5ac19ac5 but with a category cache in `CategoryFactory`.
* Serialization of node arguments is broken. For example, in pong there's an `on_body_entered` block with a subsequent `body.is_in_group('something')` test in a conditional. The `body` is being lost and I don't really understand how that works. I'll keep poking at that tomorrow.

dbnicholson:
Finally really ready for review again. It's definitely broken in the intermediate commits, but I added a few commits to smooth it out. It feels pretty working to me.

addons/block_code/ui/picker/categories/category_factory.gd
dbnicholson: This and another cache that was added in BlockCanvas were removed. Now pretty much everything BlockDefinition related happens in BlocksCatalog.

addons/block_code/serialization/block_name_tree.gd
dbnicholson: I actually like this and think it's an improvement. In the current scenario, if you move a stack of blocks, you'll still get a change because the root block position changed. And there's more in the serialization because every block has a position even if it's actually unused in all but the root block.

addons/block_code/serialization/value_block_serialization.gd
```diff
@@ -0,0 +1,9 @@
+extends Resource
+
+@export var name: StringName
+@export var arguments: Dictionary  # String, ValueBlockSerialization
```
manuq, line 4: Side note: typed dictionaries is coming to Godot 4.4, yay.

addons/block_code/ui/blocks/block/block.gd
```diff
@@ -2,55 +2,38 @@
...
-## Type of block to check if can be attached to snap point
-@export var block_type: Types.BlockType = Types.BlockType.STATEMENT
-
 ## Category to add the block to
 @export var category: String
 
+# FIXME Note: This used to be a NodePath. There is a bug in Godot 4.2 that causes the
+# reference to not be set properly when the node is duplicated. Since we don't
+# use the Node duplicate function anymore, this is okay.
+# https://github.com/godotengine/godot/issues/82670
```
manuq, line 19: :+1:

addons/block_code/ui/blocks/block/block.gd
```diff
@@ -103,42 +86,8 @@ func disconnect_signals():
 		drag_started.disconnect(c.callable)
 
 
-func update_resources(undo_redo: EditorUndoRedoManager):
```
manuq, line 118: So the new architecture makes it imposible to reuse the resources? This probably makes the resource IDs change constantly. I was hoping that we were going to drop that from William's PR. That is, have a nice AST for very well proven code generation, but keep passing the resources around.

manuq (approved):
Whoa that was a lot, thanks so much for taking care of this. I gave it a test and everything seems to work as before.

The Pong demo wasn't working because of a parse error with quoted function names: `func 'goal_right'():`, not verbatim. So I have appended a fixup commit. I only had to interact with the canvas without doing any change so the code was regenerated property.

I am a bit afraid that this drops all the efforts of having a small diff in the saved scene. Changing a single number in the canvas changes all the resources IDs. The effort was not only for the diff, reusing the existing resources was good for efficiency. I was hoping to keep that and "only" use the new AST for proper code generation. But I trust you that the other improvements are worth this regression.

addons/block_code/blocks/communication/define_method.tres
```diff
@@ -5,11 +5,12 @@
 [resource]
 script = ExtResource("1_6e473")
 name = &"define_method"
+target_node_class = ""
 description = "Define a method/function with following statements"
 category = "Communication | Methods"
 type = 1
 variant_type = 0
-display_template = "Define method {method_name: NIL}"
+display_template = "Define method {method_name: STRING_NAME}"
```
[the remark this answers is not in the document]
dbnicholson: I also tried `TYPE_CALLABLE` since that's closer to what we're actually doing in this case. You can use a dummy callable like `Callable.create(null, 'my_method')` and later get the name out with `callable.get_method()`. Unfortunately, godot only supports serializing an empty `Callable()`, so the name would be lost in the serialization.

I tried hard to think of different alternatives, but I couldn't think of anything if we are completely in the world that inputs are encoded as `Variant` and only that is used to determine the type.

addons/block_code/code_generation/blocks_catalog.gd
```diff
@@ -94,6 +94,40 @@ static func _setup_definitions_from_files():
 		_by_class_name[target][block_definition.name] = block_definition
 
 
+static func _add_output_definitions(definitions: Array[BlockDefinition]):
```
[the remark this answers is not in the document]
dbnicholson: This is actually for the Area2D and RigidBody2D `on_{entered,exited}` blocks, too. I called them output definitions because they're associated to what the code currently calls `ParameterOutput`. Those are encoded in `display_template` as `[param: TYPE]`. On the other hand, `ParameterInput` is used when the template has `{param: TYPE}`.

Technically, this is the difference between [a parameter and an argument](https://www.educative.io/answers/parameter-vs-argument). What we're trying to handle here are parameters - the things in the function declaration. What we normally use are arguments, the things used in a function call.

So, in the short term it would probably be better to call this `_add_parameter_definitions` and ignore the confusion between `ParameterInput` and `ParameterOutput`. In the longer term it would be better to change `ParameterInput` to `Argument` and `ParameterOutput` to `Parameter`. But I suspect there's a lot of loose "parameter" usage within the code.

addons/block_code/ui/blocks/block/block.gd
dbnicholson: Ah, it does regenerate IDs frequently. I hadn't paid attention to that part of the changes. I'll see if I can keep that part.

addons/block_code/blocks/communication/define_method.tres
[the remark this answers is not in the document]
dylanmccall: I haven't read through the commits here so sorry if I'm missing details, but I wonder if the thing we need in the end here is really more in the code template end of things? We need a way to say "use this parameter as written, but don't quote it or coerce it into anything new". Could communicate that like `{{method_name}}` instead of `{method_name}`.

I didn't touch that, but I _was_ poking around here in #206 by removing the pretend `OPTION` parameter type and (as a result) using NIL as a type wherever we had used `OPTION`.

dylanmccall (approved):
This all looks fine to me. This particular set of changes might have worked out better if we'd scrapped the whole repo and iterated from a clean slate (which is effectively what the final result _is_), but here we are ;)

Thank you for working through this! These are really nice improvements. I'll rebase #206 on top of this so don't feel like you need to merge for that pull request's sake. I don't see anything that urgently needs fixing except if you can look into reusing resources, but I think it's fine if the main branch just has that issue for a while. We aren't making any promises about clean git history yet.

manuq:
> This all looks fine to me.

Merging then!

addons/block_code/blocks/communication/define_method.tres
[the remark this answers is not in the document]
dbnicholson: The problem isn't so much in the template side. We can put whatever we want there and add something special. The problem comes in the generation side because everything is introspected based on the variant type. Even right now where this all happens in `ParameterInput.get_string`, everything happens based on `variant_type` except for the oddball case for `OPTION`. Actually, that part [looks pretty much the same](https://github.com/endlessm/godot-block-coding/blob/d8dae7e2f1f33248f5c039f2d03e401ae06c43bd/addons/block_code/code_generation/block_ast.gd#L139). The only reason `define_method` worked before this is that `EntryBlock` is treated specially, and [`get_entry_statement`](https://github.com/endlessm/godot-block-coding/blob/v0.6.2/addons/block_code/ui/blocks/entry_block/entry_block.gd#L22) doesn't go through the normal substitution path.

I guess a better way I hadn't thought of before is to use a dedicated class and then special case it within `TYPE_OBJECT` like `OptionData` is. E.g., `FunctionParameter` or something like that.

PR #155, opened 2024-07-23
First import fixes

manuq:
https://phabricator.endlessm.com/T35574


addons/block_code/drag_manager/drag.gd
```diff
@@ -0,0 +1,195 @@
+extends Control
```
wjt, line 1: Does this not want:

```gdscript
@tool
class_name Drag
```
?

addons/block_code/ui/blocks/utilities/drag_drop_area/drag_drop_area.gd
```diff
@@ -1,5 +1,4 @@
 @tool
-class_name DragDropArea
```
wjt, line 2: What's the effect of this change – does it prevent the class being visible to the end user?

addons/block_code/drag_manager/drag.gd
manuq: Oh good catch!

addons/block_code/ui/blocks/utilities/drag_drop_area/drag_drop_area.gd
manuq: With this, the class is not registered by name in ClassDB and it can't be referenced directly in other scripts, it must be preloaded. Also, if the script extends Node or a subclass of Node, it won't appear in the UI to the end user when adding a node.

manuq: Another advantage: it won't collide with classes defined by the end user in the project. Our current Types class has a good chance to conflict given the generic name.

manuq: I'll add this explanation to the commit message.

addons/block_code/drag_manager/drag.gd
manuq: Exactly, I just added a fixup.

dbnicholson:
FWIW, I wrote a very similar series in https://github.com/endlessm/godot-block-coding/tree/T35494-godot-4.3-importing last week. With that I was able to import on 4.3 with no warnings or errors.

addons/block_code/drag_manager/drag.gd
```diff
@@ -0,0 +1,197 @@
...
+const DragManager = preload("res://addons/block_code/drag_manager/drag_manager.gd")
+const Types = preload("res://addons/block_code/types/types.gd")
+
+var _block: Block
+var _block_scope: String
+var _block_canvas: BlockCanvas
+var _preview_block: Control
+var _snap_points: Array[Node]
+var _delete_areas: Array[Rect2]
+var action: DragManager.DragAction:
```
dbnicholson, line 14: Since `DragAction` is only used in this class, you can move the `enum` here and avoid a circular dependency between `DragManager` and `Drag`.

addons/block_code/inspector_plugin/block_script_inspector.gd
```diff
@@ -1,5 +1,7 @@
 extends EditorInspectorPlugin
 
+const BlockCodePlugin = preload("res://addons/block_code/block_code_plugin.gd")
```
dbnicholson, line 3: The reason for this is actually pretty ugly. It's to get access to the `BlockCodePlugin.main_panel` static singleton. It would be better to pass in the handler so that's not needed. I started working on it but set it aside for the moment.

manuq:
> FWIW, I wrote a very similar series in https://github.com/endlessm/godot-block-coding/tree/T35494-godot-4.3-importing last week. With that I was able to import on 4.3 with no warnings or errors.

Ouch! I didn't saw that branch, sorry. Which one would you like to favour?

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -158,7 +160,7 @@ static func get_categories(blocks: Array[Block], extra_categories: Array[BlockCa
 	# Accessing a static Callable from a static function fails in 4.2.1.
 	# Use the fully qualified name.
 	# https://github.com/godotengine/godot/issues/86032
-	cats.sort_custom(CategoryFactory._category_cmp)
+	cats.sort_custom(_category_cmp)
```
dbnicholson, line 163: This invalidates the comment above it. If I read https://github.com/godotengine/godot/issues/86032, I think this will fail on 4.2.1 with an anonymous class. It would be worth testing in 4.2.1. Then I think either we keep the global `CategoryFactory` class or change the requirements to 4.2.2.

dbnicholson:
> > FWIW, I wrote a very similar series in https://github.com/endlessm/godot-block-coding/tree/T35494-godot-4.3-importing last week. With that I was able to import on 4.3 with no warnings or errors.
> 
> Ouch! I didn't saw that branch, sorry. Which one would you like to favour?

Oh, this is fine. I just pushed it this morning when I saw this PR. I had planned to poke at it a bit more, but since you've got it going here we might as well finish it off. I would consider cherry picking [this](https://github.com/endlessm/godot-block-coding/commit/29ab2efb3c192f4b42e33e7a0760290fd726e50f). Also, as mentioned in a comment above, in [my version of breaking out the Drag class](https://github.com/endlessm/godot-block-coding/commit/5fb74eb235c8a14bb9502e22263083341d15938d), I also moved the `enum DragAction` to break a circular dependency between `DragManager` and `Drag`.

I can split out the changes to build `SimpleCharacter` and `SimpleScoring` without a scene file to a separate PR.

addons/block_code/ui/picker/categories/category_factory.gd
manuq: Ohh, I'm blind today. Let's keep the global class for now, with a comment.

manuq:
@dbnicholson thanks! Cherry-picked the instructions tree change. Tested everything again by copying to a new project and the plugin enables fine.

dbnicholson (changes requested):
Ugh. Apparently GUT does not exit with a failure on syntax errors:

```
[INFO]:  using [/home/runner/.local/share/godot/app_userdata/Block Coding Plugin] for temporary output.
Godot version:  4.2.2
GUT version:  9.2.1
SCRIPT ERROR: Parse Error: Could not find type "BlockCategory" in the current scope.
          at: GDScript::reload (res://tests/test_category_factory.gd:12)
ERROR: Failed to load script "res://tests/test_category_factory.gd" with error "Parse error".
   at: load (modules/gdscript/gdscript.cpp:2726)
[WARNING]:  Ignoring script res://tests/test_category_factory.gd because it does not extend GutTest


res://tests/test_instruction_tree.gd
* test_single_node_text
* test_root_depth_text
* test_child_node_text
* test_sibling_node_text
* test_tree_node_text
* test_script_no_nodes
* test_script_no_entry_blocks
* test_basic_script
SCRIPT ERROR: Invalid call. Nonexistent function 'insert_snapped_block' in base 'Nil'.
          at: test_basic_script (res://tests/test_instruction_tree.gd:[13](https://github.com/endlessm/godot-block-coding/actions/runs/10062323917/job/27814558872?pr=155#step:5:14)4)
    [Orphans]:  16 new orphans in test.
* test_multiple_entry_script
SCRIPT ERROR: Invalid call. Nonexistent function 'insert_snapped_block' in base 'Nil'.
          at: test_multiple_entry_script (res://tests/test_instruction_tree.gd:162)
    [Orphans]:  16 new orphans in test.
* test_signal_script
SCRIPT ERROR: Invalid call. Nonexistent function 'insert_snapped_block' in base 'Nil'.
          at: test_signal_script (res://tests/test_instruction_tree.gd:195)
    [Orphans]:  [14](https://github.com/endlessm/godot-block-coding/actions/runs/10062323917/job/27814558872?pr=155#step:5:15)5 new orphans in test.
[Orphans]:  177 new orphans in script.
10/10 passed.
```

I think all the `Nonexistent function 'insert_snapped_block' in base 'Nil'.` are due to reverting c59e48e85d6287dbb00b1c2c75e34555ad640347 but not adjusting the tests. That can be fixed separately. However, `Could not find type "BlockCategory" in the current scope.` is almost certainly caused by these changes and should be fixed. I'll try to look at making sure GUT fails in these scenarios.

manuq:
@dbnicholson ouch, so the CI was giving a false positive. Let me fix the one corresponding to this PR.

dbnicholson:
> @dbnicholson ouch, so the CI was giving a false positive. Let me fix the one corresponding to this PR.

I found https://github.com/bitwes/Gut/issues/210 upstream.

dbnicholson (approved):
Looks good now. I'll take a look at the SnapPoint fix separately.

PR #157, opened 2024-07-23
Block: Add setter to manage bottom_snap

dbnicholson:
Add a setter for bottom_snap_path so that bottom_snap can be set immediately without waiting for the node to become ready. This simplifies testing as the Block doesn't need to be added to a scene to work correctly.

https://phabricator.endlessm.com/T35573


manuq (approved):
Excellent, I was wondering how to fix the test and this is how.

PR #160, opened 2024-07-24
Text validation

manuq:
Add validation for the input text, and default values for blocks. Goal is to prevent generated code with syntax errors. It also improves the UX.

https://phabricator.endlessm.com/T35562


manuq:
The tests encountered a bug. This is now ready for review again.

dylanmccall (approved):
No complaints from me! It's great to have this :)

PR #161, opened 2024-07-26
category_factory: Add pause, stop and continue audio feature

starnight:
https://phabricator.endlessm.com/T35582


addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -492,6 +492,51 @@ static func get_general_blocks() -> Array[Block]:
...
 	block_list.append(b)
+
+	b = BLOCKS["statement_block"].instantiate()
+	b.block_name = "pause_sound"
+	b.block_type = Types.BlockType.EXECUTE
+	b.block_format = "Pause the sound {name: STRING}"
+	b.statement = (
+		"""
+		var __sound_node = get_node({name})
+		__sound_node.stream_paused = true
```
manuq, line 503: I wonder if we can follow the property setter/getter as returned by the `static func property_to_blocklist()` here, something like:

- Set sound Stream Paused to [True/False] (statement block)
- sound Stream Paused (parameter block)

This is to be closer to what GDScript does. See how Node "visible" property is handled, it is under "Graphics | Visibility" category.

In the future we should decide if we want to move the audio blocks to belong only to AudioStreamPlayer2D nodes only. Like the AnimationPlayer blocks.

manuq (changes requested):

addons/block_code/ui/picker/categories/category_factory.gd
starnight: Good think!  However, it needs to move all `#region Sounds` into `get_built_in_blocks()` which breaks the ABI.  Because, it needs an AudioStreamPlayer node first.  If we want to break the ABI in this release, I can do it.

Another easy way, like "animationplayer_play" block, have a parameter for selecting "Pause"/"Continue".  I can modify the commit in this way first.

manuq: > Good think! However, it needs to move all `#region Sounds` into `get_built_in_blocks()` which breaks the ABI. Because, it needs an AudioStreamPlayer node first. If we want to break the ABI in this release, I can do it.

Yes, I don't see why not, we are constantly breaking the format as this is still in early stage.

> Another easy way, like "animationplayer_play" block, have a parameter for selecting "Pause"/"Continue". I can modify the commit in this way first.

OK!

manuq (approved):

PR #163, opened 2024-07-26
category_factory: Tweak tooltips

cassidyjames:
"The following" was confusing to me; this makes it more clear by saying "Attached blocks"


manuq (approved):
Yes!

PR #164, opened 2024-07-26
T35536 decouple mid-term option

manuq:
Hey team, here is my alternative to the UI/data decoupling. I avoided refactors and did the minimum changes to move the block properties that don't change (like label) to its own model class.

@wnbaum maybe this can be an inbetween for the AST separation you are working on? 

https://phabricator.endlessm.com/T35536


wnbaum (approved):
Tested on my end, and everything is working the same, though the block scripts in the `pong_game` example scene should probably be resaved since the block trees are not loaded in the proper locations.

Interesting use of `StringName`, for the block name property, I didn't know it was much more performant for comparisons and dictionary use, which is used a lot in my PR - we'll have to switch it over.

I also like how you are instantiating the block scene based on the data model, something that I also do in my PR.

This is great and it does a lot for only a little refactoring. I still think it will be very useful to have the code generation and serialization options from my PR, since the structure (code blocks and indentation) of the script is still technically defined by the UI here.

I'm wondering though - all of the benefits implemented here are also present in my PR. If we merge this, I'll either have to replace a lot of your code with mine, or try to integrate your model into mine (which I think will prove difficult). Is it really worth it to have this intermediate step then? I'm a little unclear of the benefits given I've already implemented a lot of it.

manuq:
> Tested on my end, and everything is working the same, though the block scripts in the `pong_game` example scene should probably be resaved since the block trees are not loaded in the proper locations.

Good point, Will! I will update the Pong example.

> Interesting use of `StringName`, for the block name property, I didn't know it was much more performant for comparisons and dictionary use, which is used a lot in my PR - we'll have to switch it over.
> 
> I also like how you are instantiating the block scene based on the data model, something that I also do in my PR.

Yes, the only small difference is that you went for a Resouce for the model, and I thought a RefCounted would be enough (after looking at BlockCategory from @dbnicholson). What would be the benefit of using resouce for the model?

> This is great and it does a lot for only a little refactoring. I still think it will be very useful to have the code generation and serialization options from my PR, since the structure (code blocks and indentation) of the script is still technically defined by the UI here.
> 
> I'm wondering though - all of the benefits implemented here are also present in my PR. If we merge this, I'll either have to replace a lot of your code with mine, or try to integrate your model into mine (which I think will prove difficult). Is it really worth it to have this intermediate step then? I'm a little unclear of the benefits given I've already implemented a lot of it.

Considering that the primary goal is to decouple the model from the UI to allow more flexibility in blocks (localization, method overloading, etc), I think this is the wanted part. On top of the decoupling, you are doing interesting R&D and architectural changes which IMO would be better to compare in a separate step. That way it will be clear what are the benefits of having a parallel AST, how the sync with the UI tree is handled, if the stored resources are constantly changing or being reused, etc.

manuq:
I have rebased this with main branch now. But I couldn't update the Pong demo yet because of https://github.com/endlessm/godot-block-coding/issues/172 . We'll have to fix that one first.

dbnicholson:
As a minor refactor I like this a lot. The model `Block` is a bit more of the "pure data" object while the UI `Block` is more clearly a UI node with `SerializedBlockNode` bridging them. As I mentioned over in #147, naming everything `Block*` is fairly confusing. It would be nice if we could settle on some nomenclature over the 3 states - UI, code and serialized.

I'm a bit torn here. I feel like the endpoint of #147 is ultimately what I want to get to as it has a clearer separation with the AST as the intermediate state between the UI and the serialization. The downside is that it's a big set of changes. While there are parts of this PR I like, I'm a little worried that it will be mostly thrown away in #147 with extra busy work.

One thing that happened here and in #147 is that the way block elements are generated was changed. While both methods are an improvement and maybe required to rework the data model, it's a lot of churn. I don't feel strongly about either method (instantiate and modify an object vs serialized resources), but it would be good to pick one way to do it.

manuq:
We have decided to go with this after adjusting a couple things so #147 can be rebased on top:
1. Switch from RefCounted to Resource for the blocks definition.
2. Rename things properly to avoid confusion.

I'll change this to draft until is ready for review.

addons/block_code/blocks_catalog.gd
```diff
@@ -0,0 +1,38 @@
+extends Object
+
+const BlockDefinition = preload("res://addons/block_code/block_definition.gd")
+const Types = preload("res://addons/block_code/types/types.gd")
+
+static var _created: bool = false
+static var _catalog: Dictionary
```
dbnicholson, line 7: I think since the dictionary defaults to empty, the `_created` boolean is not needed. You can just check `if _catalog: return`.

manuq:
Making this draft until we release.

addons/block_code/blocks_catalog.gd
manuq: Addressed! Thanks.

manuq:
I just rebased this with main, ready for review. The renames on top are in #178

addons/block_code/ui/block_canvas/serialized_block_tree_node.gd
```diff
@@ -1,10 +1,16 @@
 class_name SerializedBlockTreeNode
 extends Resource
 
-@export var serialized_block: SerializedBlock
+@export var name: StringName
+@export var position: Vector2
```
wnbaum, line 5: Another thing here: Why does every `SerializedBlockTreeNode` have its position serialized? Only the top level block really needs its position saved. Anyways, it still works fine, with nested blocks having position (0,0), and I wouldn't change it here unless you really want to because it allows you to make very minimal changes to `BlockCanvas`.

wnbaum (approved):
Everything is looking great here!

One thing you could do is load the blocks in the blocks catalog from files since they are resources now, but this is also done in my PR. Also, I feel like it would make sense to port over some methods from CategoryFactory to the blocks catalog script, since something like `get_builtin_blocks()` would belong better in something like that.

Anyways, this is all stuff I can do in my PR. The data model is a lot closer to where we want it with a small amount of changes. Let me know if you want to make those changes here but I think this is ready to merge @manuq.

addons/block_code/ui/block_canvas/serialized_block_tree_node.gd
manuq: > Another thing here: Why does every `SerializedBlockTreeNode` have its position serialized? Only the top level block really needs its position saved. Anyways, it still works fine, with nested blocks having position (0,0), and I wouldn't change it here unless you really want to because it allows you to make very minimal changes to `BlockCanvas`.

Yes, I thought about this. For a snapped/attached block, as soon as the user detaches it by dragging it to the canvas, it will need a position. So I thought that it is better to keep the position as property to all the saved blocks, and just ignore it if they are attached. What do you think?

manuq:
> Everything is looking great here!
> 
> One thing you could do is load the blocks in the blocks catalog from files since they are resources now, but this is also done in my PR. Also, I feel like it would make sense to port over some methods from CategoryFactory to the blocks catalog script, since something like `get_builtin_blocks()` would belong better in something like that.
> 
> Anyways, this is all stuff I can do in my PR. The data model is a lot closer to where we want it with a small amount of changes. Let me know if you want to make those changes here but I think this is ready to merge @manuq.

Yeah, feel free to move the definition to separate `.tres` files in your PR, as you have the code snippet to do it already. And yes, several methods should be moved from the UI side (CategoryFactory) to the blocks catalog. You can move them as you need to reach the code generation decoupling. I don't want to block your work any more!

addons/block_code/ui/block_canvas/serialized_block_tree_node.gd
wnbaum: Yep, that's totally fine. It makes the code a little simpler in this case!

wnbaum:
> Yeah, feel free to move the definition to separate .tres files in your PR, as you have the code snippet to do it already. And yes, several methods should be moved from the UI side (CategoryFactory) to the blocks catalog. You can move them as you need to reach the code generation decoupling. I don't want to block your work any more!

Sounds good! Seems like you accomplished what we wanted to do here, so I'm going to merge!

manuq:
@wnbaum I forgot to tell you that this one had a fixup commit that needed to be squashed before merging. Thanks for the review & merge.

Issue #162, opened 2024-07-26
Pressing delete when block is focused deletes the BlockCode node


cassidyjames:
### What Happened

If I select a block and hit <kbd>Del</kbd>, I'm prompted to delete the BlockCode node in the scene tab, instead.

This was raised in the [Gamesfromscratch video around 7:05](https://youtu.be/U7n7gJIkbk0?si=3CGz3uNL2sUkzsQr&t=425), and I've confirmed it on the latest version from the AssetLib.

### How to Reproduce

1. Drag a block out to the canvas
2. Change your mind! Hit <kbd>Del</kbd>
3. See the prompt to delete the BlockCode node instead of the block

### Block Coding Plugin Version

v0.6.1 from AssetLib

### Godot Version

v4.2.2.stable.flathub [15073afe3]

### Operating System

Fedora Silverblue 40

dbnicholson:
Ouch. Possibly all we need to do is [set the focus mode](https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-property-focus-mode) on the block nodes so that they get the focus automatically. Or we can explicitly [grab the focus](https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-method-grab-focus) for the block node that's being moved around.

dbnicholson:
No, you actually need to add a handler for `KEY_DELETE`. I think I have something working, although I'm a little leery of the focusing in combination with e178d97bac141f1dc3cc30d6648b3c4224487ee8.

manuq:
It would be nice if this is accompanied with visual feedback for the focused block, the one that will be deleted. For reference, MS Arcade has a yellow outline. This is me clicking through 3 different blocks in Arcade and finally hitting DEL on them:
[Grabación de pantalla desde 2024-07-29 14-24-35.webm](https://github.com/user-attachments/assets/454c97ce-c05f-421c-af09-f787e21bd6e9)

dbnicholson:
> It would be nice if this is accompanied with visual feedback for the focused block, the one that will be deleted. For reference, MS Arcade has a yellow outline. This is me clicking through 3 different blocks in Arcade and finally hitting DEL on them: [Grabación de pantalla desde 2024-07-29 14-24-35.webm](https://github.com/user-attachments/assets/454c97ce-c05f-421c-af09-f787e21bd6e9)

[...]

PR #170, opened 2024-07-27
Block: Allow deleting with Delete button

dbnicholson:
Allow a Block to be deleted with a confirmation dialog when the Delete
key is pressed. However, when the Block is in the picker (rather than
the canvas), it shouldn't be deleted. Add a flag to disable deleting and
restore it when a drag completes. Furthermore, when it shouldn't be
deleted, stop propagating the Delete key so it doesn't try to delete the
whole BlockCode node.

Fixes: #162

### DIFF
--- a/addons/block_code/drag_manager/drag.gd
+++ b/addons/block_code/drag_manager/drag.gd
@@ -34,7 +34,7 @@ var target_snap_point: SnapPoint:
 
 var snap_block: Block:
 	get:
-		return target_snap_point.get_parent_block() if target_snap_point else null
+		return BlockTreeUtil.get_parent_block(target_snap_point) if target_snap_point else null
 
 
 func _init(block: Block, block_scope: String, offset: Vector2, block_canvas: BlockCanvas):
--- a/addons/block_code/drag_manager/drag_manager.gd
+++ b/addons/block_code/drag_manager/drag_manager.gd
@@ -80,6 +80,7 @@ func drag_ended():
 
 	if block:
 		connect_block_canvas_signals(block)
+		block.grab_focus()
 
 	_block_canvas.release_scope()
 
--- a/addons/block_code/drag_manager/drag_manager.gd
+++ b/addons/block_code/drag_manager/drag_manager.gd
@@ -82,6 +82,9 @@ func drag_ended():
 		connect_block_canvas_signals(block)
 		block.grab_focus()
 
+		# Allow the block to be deleted now that it's on the canvas.
+		block.can_delete = true
+
 	_block_canvas.release_scope()
 
 	drag.queue_free()
--- a/addons/block_code/ui/block_tree_util.gd
+++ b/addons/block_code/ui/block_tree_util.gd
@@ -78,3 +78,11 @@ static func get_tree_scope(node: Node) -> String:
 		if scope != "":
 			return scope
 	return ""
+
+
+## Get the nearest Block node that is a parent of the provided node.
+static func get_parent_block(node: Node) -> Block:
+	var parent = node.get_parent()
+	while parent and not parent is Block:
+		parent = parent.get_parent()
+	return parent as Block
--- a/addons/block_code/ui/blocks/block/block.gd
+++ b/addons/block_code/ui/blocks/block/block.gd
@@ -48,6 +48,7 @@ func _set_bottom_snap_path(value: NodePath):
 func _ready():
 	if bottom_snap == null:
 		_set_bottom_snap_path(bottom_snap_path)
+	focus_mode = FocusMode.FOCUS_ALL
 	mouse_filter = Control.MOUSE_FILTER_IGNORE
 
 
--- a/addons/block_code/ui/blocks/block/block.gd
+++ b/addons/block_code/ui/blocks/block/block.gd
@@ -39,6 +39,9 @@ signal modified
 # https://github.com/godotengine/godot/issues/82670
 var bottom_snap: SnapPoint
 
+## Whether the block can be deleted by the Delete key.
+var can_delete: bool = true
+
 
 func _set_bottom_snap_path(value: NodePath):
 	bottom_snap_path = value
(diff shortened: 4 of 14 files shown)


dbnicholson:
@dylanmccall I'd appreciate if you could take a look at the mouse focusing since I'm not totally confident about how it interacts with the `mouse_filter` addition in e178d97bac141f1dc3cc30d6648b3c4224487ee8.

Another thing that would be super useful (but I didn't attempt to tackle) is if the block could be drawn to show that it's focused. See [the docs](https://docs.godotengine.org/en/stable/tutorials/ui/custom_gui_controls.html#checking-focus) for a description. However, it's a bit difficult in our case since the thing we want to alter is the background node within the Block.

manuq:
I don't think a confirmation is needed, since it's a bit cumbersome and its already easy to undo. For referece, MS Arcade has none.

Maybe deleting a block should not delete the ones stacked to its bottom?

Regarding focus, something is wrong in both main and this branch: when I scroll in the Output dock the Block Code canvas is zoomed in and out. As you can see in this video:
[Grabación de pantalla desde 2024-07-29 14-44-24.webm](https://github.com/user-attachments/assets/3c4fe60d-87ac-4ce7-b1c9-7178c1801c2b)

dbnicholson:
> I don't think a confirmation is needed, since it's a bit cumbersome and its already easy to undo. For referece, MS Arcade has none.

I guess I followed the scene editor where it asks for confirmation when deleting a node. Personally I find it helpful but I'm mostly indifferent.

> Maybe deleting a block should not delete the ones stacked to its bottom?

I also was following the scene editor where it will delete the whole branch. Although the dialog does say "Delete ... and all it's children" in that case. Reparenting the lower blocks might be a little tricky. I'm also not sure if that could leave you with an invalid node tree. I'm inclined to delete the whole branch like the scene editor and leave any fancier work for a followup.

> Regarding focus, something is wrong in both main and this branch: when I scroll in the Output dock the Block Code canvas is zoomed in and out. As you can see in this video: [Grabación de pantalla desde 2024-07-29 14-44-24.webm](https://github.com/user-attachments/assets/3c4fe60d-87ac-4ce7-b1c9-7178c1801c2b)

Oh, interesting. I also noticed that my block canvas was way zoomed in at some point but hadn't made the connection. I guess I need to explicitly [[ https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-method-release-focus | release focus ]] when the canvas is no longer active.

manuq:
> > I don't think a confirmation is needed, since it's a bit cumbersome and its already easy to undo. For referece, MS Arcade has none.
> 
> I guess I followed the scene editor where it asks for confirmation when deleting a node. Personally I find it helpful but I'm mostly indifferent.

Ok! I checked visual shader nodes and also don't have a confirmation. But I think we need learners feedback. So feel free to merge with the confirmation. Or maybe @cassidyjames or @jbourqueendless can tell us.
[Grabación de pantalla desde 2024-07-30 16-22-46.webm](https://github.com/user-attachments/assets/98617e5d-f4cf-4ed3-bc9f-8f08a2e265e7)


> > Maybe deleting a block should not delete the ones stacked to its bottom?
> 
> I also was following the scene editor where it will delete the whole branch. Although the dialog does say "Delete ... and all it's children" in that case. Reparenting the lower blocks might be a little tricky. I'm also not sure if that could leave you with an invalid node tree. I'm inclined to delete the whole branch like the scene editor and leave any fancier work for a followup.

Fair! I wasn't thinking them as children but as "stacked below". Like removing a line of code should not remove the ones below.

> > Regarding focus, something is wrong in both main and this branch: when I scroll in the Output dock the Block Code canvas is zoomed in and out. As you can see in this video: [Grabación de pantalla desde 2024-07-29 14-44-24.webm](https://github.com/user-attachments/assets/3c4fe60d-87ac-4ce7-b1c9-7178c1801c2b)
> 
> Oh, interesting. I also noticed that my block canvas was way zoomed in at some point but hadn't made the connection. I guess I need to explicitly [[ https://docs.godotengine.org/en/stable/classes/class_control.html#class-control-method-release-focus | release focus ]] when the canvas is no longer active.

:+1:

dylanmccall:
I played with this a bunch today, and created #179 adding focus indicators. Feel free to merge it into this branch if it's easier to deal with :) It comes with the bonus that we almost have working keyboard navigation, although I'm very fed up with Godot's theming system at the moment. Need to sit down and try to understand it. Right now I'm doing a weird thing where I call `add_theme_stylebox_override` with a new "panel" stylebox when the focus changes, which doesn't seem right. It seems like there _should_ be a "normal" and "focus" stylebox, but that isn't what's happening.

dbnicholson:
> Regarding focus, something is wrong in both main and this branch: when I scroll in the Output dock the Block Code canvas is zoomed in and out. As you can see in this video: [Grabación de pantalla desde 2024-07-29 14-44-24.webm]

I can't reproduce this anymore either in main, the branch here (based on an older main), or a rebased version of this branch. What I started looking at was disabling input event handling for the block canvas when the main panel is hidden:

```
diff --git a/addons/block_code/ui/main_panel.gd b/addons/block_code/ui/main_panel.gd
index ce628ce..39504ed 100644
--- a/addons/block_code/ui/main_panel.gd
+++ b/addons/block_code/ui/main_panel.gd
@@ -47,6 +47,8 @@ func _ready():
        if not _collapse_button.icon:
                _collapse_button.icon = _icon_collapse
 
+       visibility_changed.connect(_on_visibility_changed)
+
 
 func _on_undo_redo_version_changed():
        if _current_block_code_node == null:
@@ -79,6 +81,11 @@ func _on_delete_node_button_pressed():
        pass  # Replace with function body.
 
 
+func _on_visibility_changed():
+       print("Main panel visibility changed to %s" % visible)
+       _block_canvas.set_process_input(visible)
+
+
 func _on_delete_dialog_confirmed(block_code_node: BlockCode):
        var parent_node = block_code_node.get_parent()
 
```

Since I couldn't reproduce it anymore, I decided to leave that out.

dbnicholson:
Ready for review again. It still has the confirmation dialog and still deletes the entire tree of blocks below it. What's changed:

* I pulled in #179 with [a minor change](https://github.com/endlessm/godot-block-coding/pull/179/files#r1700754502).
* I went with @wnbaum's suggestion to use a simpler `Delete <N> blocks?` prompt like in Scratch. I really wanted to put a useful name there but we don't have anything at the moment and I didn't want to go making a bunch of changes in `CategoryFactory` when we're about to redo that whole thing.

dylanmccall:
I gave this another look and I'm happy with where it is. I _was_ able to reproduce that issue with scrolling in other bottom panel tabs, but since it isn't a regression in this branch, I reported it separately in #184. I rebased this branch against main, and I'll merge it in a moment :)

dylanmccall (approved):

PR #173, opened 2024-07-30
Parameter input: Fix Vector2 split

manuq:
If a space is left in the Y axis, the default doesn't work. This fixes a bug where focusing the Y side of a Vector2 input and then losing the focus ends with the " " string in the input, losing the default number.

Follow up of https://github.com/endlessm/godot-block-coding/pull/160

https://phabricator.endlessm.com/T35562

### DIFF
--- a/addons/block_code/ui/picker/categories/category_factory.gd
+++ b/addons/block_code/ui/picker/categories/category_factory.gd
@@ -576,14 +576,14 @@ static func property_to_blocklist(property: Dictionary) -> Array[Block]:
 	const FALLBACK_SET_FOR_TYPE = {
 		TYPE_INT: "0",
 		TYPE_FLOAT: "0",
-		TYPE_VECTOR2: "0, 0",
+		TYPE_VECTOR2: "0,0",
 		TYPE_COLOR: "DARK_ORANGE",
 	}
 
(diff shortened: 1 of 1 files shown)


wnbaum (changes requested):
I think you also need to add a space after the comma in [this](https://github.com/endlessm/godot-block-coding/pull/173/files#diff-6d1ff8af3933120fff67da3dc102bea079fc558502f4bae6701e7b69876e3377R79) line. Otherwise when the script is loaded again, the vector string will not be split properly.

You'll need to update the pong scene as well, since vectors are currently stored like `"1,1"` without the space.

Another option is just to remove the spaces entirely from the default values to keep the current serialization scheme.

manuq:
Thanks for the insightful review, @wnbaum! As you said, the problem was actually in the defaults having the space character after the comma. Ready for review again.

wnbaum (approved):
Looks good to me now!

Having the user set properties serialized as a String is kind of weird, I changed this in my decouple PR so that Variants themselves get serialized which hopefully leads to less issues like this.

dbnicholson:
> Having the user set properties serialized as a String is kind of weird, I changed this in my decouple PR so that Variants themselves get serialized which hopefully leads to less issues like this.

I was thinking the same thing. Let's try to get that in when the dust has settled.

Issue #172, opened 2024-07-29
Wrong code generation with two print blocks


manuq:
### What Happened

Code generation is not working as expected with stacked blocks after the scene is reloaded. It seems that something is wrong with the bottom snap.

### How to Reproduce

1. Add a BlockCode node to any node
2. Attach 2 Print blocks to the Ready block, change the text to "one" and "two" respectively
3. Print the generated code, it should work as expected
4. Save the scene and load it again
5. Print the generated code -> only the first print is present

### Block Coding Plugin Version

main

### Godot Engine Version

v4.2.2

### Operating System

Endless

### Logs or command-line output

```shell
None.
```

### Anything else?

Also dragging the print blocks to reorder them doesn't work. This is before saving:
[Grabación de pantalla desde 2024-07-29 17-39-32.webm](https://github.com/user-attachments/assets/bcd70313-c9d1-4392-acc1-2254a8571051)

This is after saving:
[Grabación de pantalla desde 2024-07-29 17-33-18.webm](https://github.com/user-attachments/assets/4a4fd411-4bfd-46c0-bb60-e57b7aeb1a7e)

PR #175, opened 2024-07-30
Block: Set bottom_snap_path when ready

manuq:
The bottom_snap must be set on ready, regardless of previously setting it.

Regression from 9aae2bb6

Fixes: #172


dbnicholson (approved):
Thanks. I'm still getting confused on when variables are set by Godot. I suppose that what's happening is that the node hasn't been added to the scene yet, so the initial `get_node_or_null` returns `null`.

PR #176, opened 2024-07-30
Block Canvas: Reuse saved block scene resource

manuq:
By simply assigning it to the scene resource property when loading the tree.

Follow up of 01ab76c5 which already removed all resource regeneration except when loading the canvas.


manuq:
I decided to split this from #164.

A way to test this is to load the Pong example, move a block, save the scene, and then check the diff. Only the position should be the noticeable change (besides outdated format):
Captura desde 2024-07-30 16-48-06

dbnicholson (approved):
Very nice!

PR #177, opened 2024-07-31
Improve ControlBlock background drawing

dylanmccall:
Instead of reusing the background.gd script to draw a gutter in ControlBlock, define a new script (gutter.gd) specifically for that purpose. In the existing background.gd script, draw a left border which is expected to flow into the gutter if shift_top or shift_bottom is set.

https://phabricator.endlessm.com/T35538

### DIFF
--- a/addons/block_code/ui/blocks/control_block/control_block.gd
+++ b/addons/block_code/ui/blocks/control_block/control_block.gd
@@ -2,8 +2,10 @@
 class_name ControlBlock
 extends Block
 
+const Background = preload("res://addons/block_code/ui/blocks/utilities/background/background.gd")
 const Constants = preload("res://addons/block_code/ui/constants.gd")
 const DragDropArea = preload("res://addons/block_code/ui/blocks/utilities/drag_drop_area/drag_drop_area.gd")
+const Gutter = preload("res://addons/block_code/ui/blocks/utilities/background/gutter.gd")
 
 @export var block_formats: Array = []
 @export var statements: Array = []
@@ -104,9 +106,8 @@ func format():
 		row.custom_minimum_size.y = 30
 		row.size_flags_horizontal = Control.SIZE_SHRINK_BEGIN
 
-		var bg := Control.new()
+		var bg := Background.new()
 		bg.name = "Background"
-		bg.set_script(preload("res://addons/block_code/ui/blocks/utilities/background/background.gd"))
 		bg.color = color
 		if i != 0:
 			bg.shift_top = Constants.CONTROL_MARGIN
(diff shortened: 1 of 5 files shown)


addons/block_code/ui/blocks/control_block/control_block.gd
```diff
@@ -148,9 +143,18 @@ func format():
 		snap_container.custom_minimum_size.x = 30
 		snap_container.custom_minimum_size.y = 30
 		snap_container.size_flags_horizontal = Control.SIZE_SHRINK_BEGIN
-		snap_container.add_theme_constant_override("margin_left", Constants.CONTROL_MARGIN)
+
+		var snap_gutter := Control.new()
+		snap_gutter.name = "Background"
+		snap_gutter.set_script(preload("res://addons/block_code/ui/blocks/utilities/background/gutter.gd"))
```
dbnicholson, line 149: I've seen this in a couple places in here. Why not just instantiate the class directly?

dylanmccall: I'm not sure _why_ we don't do that, indeed. I'll fix that :)

manuq: Yep, `set_script()` only makes sense in the BlockCode node when it replaces the parent script. All others should be fixed.

manuq (changes requested):
It looks very nice! The `set_script` that was there before for other backgrounds and should be changed. Besides that, please leave the Pong scene out of this PR.

dylanmccall:
In the newest set of commits, I removed the unnecessary change to `pong_game.tscn` and I replaced that `set_script()` scheme with instantiating `BlockBackground` and `BlockGutter` classes directly.

addons/block_code/ui/blocks/control_block/control_block.gd
```diff
@@ -144,9 +143,8 @@ func format():
 		snap_container.custom_minimum_size.y = 30
 		snap_container.size_flags_horizontal = Control.SIZE_SHRINK_BEGIN
 
-		var snap_gutter := Control.new()
+		var snap_gutter := BlockGutter.new()
```
dbnicholson, line 146: I think we've been trying to move away from global classes unless their really needed since it reduces the chance for startup issues with the global class cache. Instead, I think you should just preload them at the top:

```
const BlockBackground = preload("res://addons/block_code/ui/blocks/utilities/background/background.gd")
const BlockGutter = preload("res://addons/block_code/ui/blocks/utilities/background/gutter.gd")
```

dbnicholson (changes requested):
I can't say that I really followed the polygon changes, but I trust their right. I just had one small suggestion on class names.

addons/block_code/ui/blocks/control_block/control_block.gd
manuq: Yes, exactly. Can you make this last change @dylanmccall ?

addons/block_code/ui/blocks/utilities/background/background.gd
```diff
@@ -1,4 +1,5 @@
 @tool
+class_name BlockBackground
```
manuq, line 2: As Dan pointed, please remove `class_name` and use preload instead.

manuq (changes requested):

addons/block_code/ui/blocks/control_block/control_block.gd
dylanmccall: Ah, I missed that happening! Makes sense :) That's fixed in the newest commit.

dbnicholson (approved):
Lovely, thanks!

addons/block_code/ui/blocks/control_block/control_block.gd
dbnicholson: I think Manuel started it near the beginning of GUADEC in the context of [import errors](https://phabricator.endlessm.com/T35494) and I happened to notice it since I was trying to wrap up some of my own work that Thursday and Friday.

PR #179, opened 2024-08-01
Add focus indicator for blocks

dylanmccall:
Note that this requires ac4c748ed8fc88768f1618c7ca227155f599890d from #170.

Helps: #162


dbnicholson:
Nice! This is exactly what I was trying to do the other day.

addons/block_code/ui/blocks/parameter_block/parameter_block.gd
```diff
@@ -16,16 +17,24 @@ var param_name_input_pairs: Array
...
 func _ready():
 	super()
 
+	_panel_normal = _panel.get_theme_stylebox("panel").duplicate()
+	_panel_normal.bg_color = color
+	_panel_normal.border_color = color.darkened(0.2)
+
+	_panel_focus = _panel.get_theme_stylebox("panel").duplicate()
+	_panel_focus.bg_color = color
+	_panel_focus.border_color = Constants.FOCUS_BORDER_COLOR
```
dbnicholson, line 33: This is indeed awkward since there are now 3 styleboxes including the one saved in the scene. Is it possible to have one stylebox in the code that duplicates the one in the scene and then just mutate the color properties on it instead of overriding back and forth? You could even go further and build the initial override in code instead of taking it from the scene, but I feel like people would prefer to edit the non-color properties in the editor.

dbnicholson (approved):
This is great! The stylebox thing is weird but I don't think there's a ton of room for improvement. I'll probably pull this into my branch as is while I address some other comments there.

addons/block_code/ui/blocks/parameter_block/parameter_block.gd
manuq: > This is indeed awkward since there are now 3 styleboxes including the one saved in the scene. Is it possible to have one stylebox in the code that duplicates the one in the scene and then just mutate the color properties on it instead of overriding back and forth? You could even go further and build the initial override in code instead of taking it from the scene, but I feel like people would prefer to edit the non-color properties in the editor.

For reference, other Control nodes in Godot have several styleboxes, one per each style. For example Button has styleboxes for Normal, Pressed, Disabled, Focus.
Captura desde 2024-08-01 10-49-20

dylanmccall: Oooh, should we attach a bunch of different styleboxes and then use `theme_type_variation` to switch between them? Right now I have it swapping the stylebox for the "Panel" style, but looking at the documentation for that property I see…

> The name of a theme type variation used by this Control to look up its own theme items. When empty, the class name of the node is used (e.g. Button for the Button control), as well as the class names of all parent classes (in order of inheritance).

… which helps me to understand why styleboxes have these two completely different naming schemes.

addons/block_code/ui/blocks/parameter_block/parameter_block.gd
```diff
@@ -73,3 +82,11 @@ static func get_scene_path():
...
 func format():
 	param_name_input_pairs = StatementBlock.format_string(self, %HBoxContainer, block_format, defaults)
+
+
+func _on_focus_entered():
+	_panel.add_theme_stylebox_override("panel", _panel_focus)
+
+
+func _on_focus_exited():
+	_panel.add_theme_stylebox_override("panel", _panel_normal)
```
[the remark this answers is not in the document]
dbnicholson: This does not appear to be the case. The scene editor doesn't seem to focus the node but rather focus the entry in the scene tree editor.

addons/block_code/ui/blocks/parameter_block/parameter_block.gd
dbnicholson: That's certainly very interesting. I suppose that would be good because we'd be inheriting whatever Panel theme variation is in the project. I'm going to punt and keep what you have for now, but feel free to keep investigating how to better manage this!

addons/block_code/ui/util.gd
```diff
@@ -12,3 +12,11 @@ static func node_is_part_of_edited_scene(node: Node) -> bool:
 
 	var edited_scene_parent := tree.edited_scene_root.get_parent()
 	return edited_scene_parent and edited_scene_parent.is_ancestor_of(node)
+
+
+## Get the nearest Block node that is a parent of the provided node.
+static func get_parent_block(node: Node) -> Block:
```
dbnicholson, line 18: I ended up putting this in the new `BlockTreeUtil` class I made so as not to re-introduce a circular dependency. The `Util` class otherwise depends only on native classes.

dbnicholson:
I've pulled this in to #170 now. Let's have any further discussion there.

PR #182, opened 2024-08-01
Fix InstructionTree and Block circular dependency

dbnicholson:
Since 4338632 there's been a circular dependency between InstructionTree and Block that manifests as a script error due to a non-existent resource in a seemingly unrelated object. Several components use InstructionTree.TreeNode including Block, and prior to 4338632 there were no references to any non-native classes. Split out the block tree functions to a new BlockTreeUtil class so that InstructionTree no longer contains references to Block.


dbnicholson:
This will almost certainly cause conflicts in the decoupling PRs.

addons/block_code/drag_manager/drag.gd
```diff
@@ -3,6 +3,7 @@ extends Control
 
 const Background = preload("res://addons/block_code/ui/blocks/utilities/background/background.gd")
 const BlockCanvas = preload("res://addons/block_code/ui/block_canvas/block_canvas.gd")
+const BlockTreeUtil = preload("res://addons/block_code/ui/block_tree_util.gd")
```
dbnicholson, line 6: I'm not at all convinced of this name or path. What seemed to be in common in the functions moved was that they were about a tree of `Block` nodes. @wnbaum's work completely changes the way this is handled.

manuq:
> This will almost certainly cause conflicts in the decoupling PRs.

No problem, we can rebase.

addons/block_code/drag_manager/drag.gd
manuq: If this is going to be changed anyways maybe the current name/path doesn't matter that much?

manuq (approved):
Excellent, this fixes the circular dependency in a decent way with a separate script. I tested it in both the development project and by copying the plugin to a new project and then enabling it. No errors present.

Issue #187, opened 2024-08-07
The zoom label is impossible to see with certain editor themes


dylanmccall:
### What Happened

Using the "Light" editor theme, the zoom label is almost impossible to see:

![image](https://github.com/user-attachments/assets/27147eb9-122f-41b6-ae98-ac5093a87e17)

### How to Reproduce

1. In Godot's Editor Settings, switch to a theme with a light background colour, such as the "Light" theme preset.
2. In the block code editor, look for the zoom label in the bottom right corner of the screen.

manuq:
Fixed in main by the PR above. Will be in next release.

Issue #186, opened 2024-08-07
It is possible to get lost in the editor when changing the zoom level


dylanmccall:
### What Happened

When using the block coding editor, it is possible to get lost, particularly when zooming as well as panning. If I find myself looking at a blank canvas, it is unclear what direction to pan in order to find my blocks again.

### How to Reproduce

When using the pong_game example, use the mouse wheel zoom out a long way, then zoom in on an empty space.

### Block Coding Plugin Version

v0.6.2

### Godot Engine Version

v4.2.2.stable

manuq:
Fixed by the PR above. Will be in next release.

PR #188, opened 2024-08-07
Zoom label enhancements

dylanmccall:
For https://github.com/endlessm/godot-block-coding/issues/186, this adds a feature which resets the editor zoom level and scroll position when the user clicks the zoom label (now button) in the bottom right corner. This is consistent with the zoom button in Godot's 2D editor, although for our purpose I felt better about _always_ resetting the scroll position.

For https://github.com/endlessm/godot-block-coding/issues/187, this changes the zoom label to use `CanvasItem.modulate` instead of overriding the text colour.


starnight (approved):
Useful features!  Thanks!

PR #191, opened 2024-08-10
Generate value blocks for AnimationPlayer animations 

dylanmccall:
https://phabricator.endlessm.com/T35564

### DIFF
--- a/addons/block_code/ui/main_panel.gd
+++ b/addons/block_code/ui/main_panel.gd
@@ -55,7 +55,7 @@ func _on_undo_redo_version_changed():
 		return
 
 	var block_script: BlockScriptSerialization = _current_block_code_node.block_script
-	_picker.block_script_selected(block_script)
+	_picker.block_script_selected(block_script, _current_block_code_node.get_parent() if _current_block_code_node else null)
 	_title_bar.block_script_selected(block_script)
 	_block_canvas.block_script_selected(block_script)
 
@@ -117,7 +117,7 @@ func switch_block_code_node(block_code_node: BlockCode):
 	_delete_node_button.disabled = _current_block_code_node == null
 	if _current_block_code_node != null:
 		_try_migration()
-	_picker.block_script_selected(block_script)
+	_picker.block_script_selected(block_script, _current_block_code_node.get_parent() if _current_block_code_node else null)
 	_title_bar.block_script_selected(block_script)
 	_block_canvas.block_script_selected(block_script)
 
--- a/addons/block_code/ui/main_panel.gd
+++ b/addons/block_code/ui/main_panel.gd
@@ -53,11 +53,7 @@ func _ready():
 func _on_undo_redo_version_changed():
 	if _current_block_code_node == null:
 		return
-
-	var block_script: BlockScriptSerialization = _current_block_code_node.block_script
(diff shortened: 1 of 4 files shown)


addons/block_code/ui/main_panel.gd
```diff
@@ -55,7 +55,7 @@ func _on_undo_redo_version_changed():
 		return
 
 	var block_script: BlockScriptSerialization = _current_block_code_node.block_script
-	_picker.block_script_selected(block_script)
+	_picker.block_script_selected(block_script, _current_block_code_node.get_parent() if _current_block_code_node else null)
```
manuq: The line above is assuming that `_current_block_code_node` is not null (it is referencing `block_script` in it.

In any case, this would be a bit more readable without a ternary operator.

addons/block_code/ui/main_panel.gd
```diff
@@ -117,7 +117,7 @@ func switch_block_code_node(block_code_node: BlockCode):
 	_delete_node_button.disabled = _current_block_code_node == null
 	if _current_block_code_node != null:
 		_try_migration()
-	_picker.block_script_selected(block_script)
+	_picker.block_script_selected(block_script, _current_block_code_node.get_parent() if _current_block_code_node else null)
```
manuq, line 121: Nit: this would be a bit more readable without a ternary operator. Or at least moving the conditional to a variable:
```
var parent_node: Node = _current_block_code_node.get_parent() if _current_block_code_node else null
_picker.block_script_selected(block_script, parent_node)
```

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -628,6 +628,11 @@ static func blocks_from_property_list(property_list: Array, selected_props: Dict
 	return block_list
 
 
+static func get_blocks_for_object(object: Object) -> Array[Block]:
+	var blocks: Array[Block] = []
+	return blocks
```
manuq: :+1:

manuq (approved):
This looks great, only a nit comment. I think it's fine to refresh the animations every time the panel is shown, as the last commit does. Maybe later we can add a refresh button in the title bar.

I wonder if we should join forces and land #147 first.

manuq (changes requested):
Sorry for the double review. Now I gave it a try and it's not working. The generated script is missing the quotes around the aniimation name:
[Grabación de pantalla desde 2024-08-12 15-22-52.webm](https://github.com/user-attachments/assets/154433af-cd1a-4db4-bbf8-7d470e05f803)

This makes me wonder if the animation values should be options in a dropdown of the Play block. Otherwise these are string blocks that can be used anywhere else, but are only relevant for playing the animation.

dylanmccall:
> This makes me wonder if the animation values should be options in a dropdown of the Play block. Otherwise these are string blocks that can be used anywhere else, but are only relevant for playing the animation.

Yeah, I'll give that a try :) In that case it'll definitely need to be on top of #147, and might need some new pieces? But I'll see where I can get it!

dbnicholson:
@dylanmccall I believe this all got reimplemented in #206, correct? Can we close this?

dylanmccall:
Yes! Thanks for remembering.

PR #194, opened 2024-08-14
Add 4 random number generation blocks

jfcharron:
Add 4 random number generation blocks.
One for each of these godot's functions:

1. [`randi`](https://docs.godotengine.org/en/stable/classes/class_%40globalscope.html#class-globalscope-method-randi)
2. `randi_range(from,to)`
3. `randf`
4. `randf_range(from,to)`

### DIFF
--- a/addons/block_code/ui/picker/categories/category_factory.gd
+++ b/addons/block_code/ui/picker/categories/category_factory.gd
@@ -198,7 +198,7 @@ static func get_general_blocks() -> Array[Block]:
 	block_list.append(b)
 
 	# Math
-	for block_name in [&"add", &"subtract", &"multiply", &"divide", &"pow"]:
+	for block_name in [&"add", &"subtract", &"multiply", &"divide", &"pow", &"randf_range", &"randi_range"]:
 		b = Util.instantiate_block(block_name)
 		block_list.append(b)
 
--- a/addons/block_code/blocks/math/randf_range.tres
+++ b/addons/block_code/blocks/math/randf_range.tres
@@ -1,0 +1,19 @@
+[gd_resource type="Resource" load_steps=2 format=3 uid="uid://u35glf576fue"]
+
+[ext_resource type="Script" path="res://addons/block_code/code_generation/block_definition.gd" id="1_03jps"]
+
+[resource]
+script = ExtResource("1_03jps")
+name = &"randf_range"
+description = "Generate a random floating point number between [i]from[/i] and [i]to[/i] inclusively"
+category = "Math"
+type = 3
+variant_type = 3
+display_template = "Random floating point number between {from: FLOAT} and {to: FLOAT}"
+code_template = "randf_range({from}, {to})"
+defaults = {
+"from": -1.0,
+"to": 1.0
+}
+signal_name = ""
+scope = ""
--- a/addons/block_code/blocks/math/randi_range.tres
+++ b/addons/block_code/blocks/math/randi_range.tres
@@ -1,0 +1,19 @@
+[gd_resource type="Resource" load_steps=2 format=3 uid="uid://b3b1dyarh2hmo"]
+
+[ext_resource type="Script" path="res://addons/block_code/code_generation/block_definition.gd" id="1_hk574"]
+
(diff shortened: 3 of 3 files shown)


addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -404,6 +404,46 @@ static func get_general_blocks() -> Array[Block]:
...
+	b.category = "Math"
+	b.tooltip_text = "Generate a random unsigned 32-bits integer number between 0 and 2^32-1 inclusively ([0, 2^32-1])"
+	block_list.append(b)
+
+	b = BLOCKS["parameter_block"].instantiate()
+	b.block_name = "randi_range"
+	b.variant_type = TYPE_INT
+	b.block_format = "Random number between {from: INT} and {to: INT}"
+	b.statement = "(randi_range({from}, {to}))"
+	b.defaults = {"from": -1, "to": 1}
```
manuq, line 442: Maybe the 2 "random integer" blocks can be one, if the defaults are `from: 0, to: 2^32-1`. Then the code (statement) can be:
```
randi() if {from} == 0 and {to} == 2^32-1 else randi_range({from}, {to})
```

addons/block_code/ui/picker/categories/category_factory.gd
```diff
@@ -404,6 +404,46 @@ static func get_general_blocks() -> Array[Block]:
...
+	b.category = "Math"
+	b.tooltip_text = "Generate a random floating point number between 0.0 and 1.0 inclusively ([0.0, 1.0])"
+	block_list.append(b)
+
+	b = BLOCKS["parameter_block"].instantiate()
+	b.block_name = "randf_range"
+	b.variant_type = TYPE_FLOAT
+	b.block_format = "Random number between {from: FLOAT} and {to: FLOAT}"
+	b.statement = "(randf_range({from}, {to}))"
+	b.defaults = {"from": -1.0, "to": 1.0}
```
manuq, line 422: Maybe the 2 "random floating point" blocks can be one, if the defaults are `from: 0.0, to: 1.0`. Then the code (statement) can be:
```
randf() if {from} == 0.0 and {to} == 1.1 else randf_range({from}, {to})
```

manuq (approved):
This looks good, thanks @jfcharron ! Can you give me permission to edit your PR? https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

So I can squash the commits and merge.

addons/block_code/ui/picker/categories/category_factory.gd
jfcharron: I will merge the two blocks, because they are pretty similar. Also I will do the same for randi.

I feel the if statement is unnecessary because after some thought, I believe that the randi method is not useful in our case. I don't see a case where you would want such a high number of random values. So I will just the let the users enter the range they desire.

jfcharron:
The `Allow edits by maintainers` checkbox is checked so I suppose you should be able to merge.

addons/block_code/ui/picker/categories/category_factory.gd
manuq: > I will merge the two blocks, because they are pretty similar. Also I will do the same for randi.
> 
> I feel the if statement is unnecessary because after some thought, I believe that the randi method is not useful in our case. I don't see a case where you would want such a high number of random values. So I will just the let the users enter the range they desire.

Yes, very well thought @jfcharron. Programming with blocks should be easy and providing multiple blocks that do the same can be confusing to the end users.

manuq:
@jfcharron merged! I tested placing a sprite in a random vertical position and it worked like a charm:
Captura desde 2024-08-15 09-07-18

I have converted the blocks to the new definition format, which uses resources. In case you want to contribute new blocks, this is now the way to do it. It is nicer because the definitions can be edited in the editor directly:
Captura desde 2024-08-15 09-25-07

Thanks!

PR #195, opened 2024-08-14
T35591 Add remaining general blocks

manuq:
This completes the port for blocks that don't depend on the node class.

Changes compared with #147:
- The definitions for sound, graphics and "await_scene_ready" blocks were missing.
- Added some missing defaults too.

https://phabricator.endlessm.com/T35591

### DIFF
--- a/addons/block_code/blocks/communication/call_method_node.tres
+++ b/addons/block_code/blocks/communication/call_method_node.tres
@@ -5,11 +5,12 @@
 [resource]
 script = ExtResource("1_pg363")
 name = &"call_method_node"
-type = 2
-variant_type = 0
-display_template = "Call method {method_name: STRING} on node {node: OBJECT}"
-code_template = "{node}.call({method_name})"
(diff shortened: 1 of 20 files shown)


dbnicholson (approved):
I didn't look super closely at every commit, but they seem correct. Since these changes are primarily copied from #147, I think it would be nice to retain @wnbaum as the author of the commits. Like `git commit --amend --author='Will Baumgartner <wnb@endlessos.org>'`.

manuq:
> I didn't look super closely at every commit, but they seem correct. Since these changes are primarily copied from #147, I think it would be nice to retain @wnbaum as the author of the commits. Like `git commit --amend --author='Will Baumgartner <wnb@endlessos.org>'`.

Good point! Although most of this is not in #147 I will add @wnbaum as co-author.

dylanmccall (approved):
Yeah, I didn't spot any issues here either :)

PR #197, opened 2024-08-15
Move development project to Godot 4.3

manuq:

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -25,6 +25,9 @@ jobs:
 
   tests:
     name: Tests
+    strategy:
+      matrix:
+        godot-version: [4.2.2, 4.3.0]
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
@@ -32,7 +35,7 @@ jobs:
     - name: Setup Godot
       uses: chickensoft-games/setup-godot@v2.1.1
       with:
-        version: 4.2.2
+        version: ${{ matrix.godot-version }}
         use-dotnet: false
     - name: Initialize Godot
       run: |
--- a/addons/gut/fonts/AnonymousPro-Bold.ttf.import
+++ b/addons/gut/fonts/AnonymousPro-Bold.ttf.import
@@ -15,6 +15,7 @@ dest_files=["res://.godot/imported/AnonymousPro-Bold.ttf-9d8fef4d357af5b52cd60af
 Rendering=null
 antialiasing=1
 generate_mipmaps=false
+disable_embedded_bitmaps=true
 multichannel_signed_distance_field=false
 msdf_pixel_range=8
 msdf_size=48
--- a/addons/gut/fonts/AnonymousPro-BoldItalic.ttf.import
+++ b/addons/gut/fonts/AnonymousPro-BoldItalic.ttf.import
@@ -15,6 +15,7 @@ dest_files=["res://.godot/imported/AnonymousPro-BoldItalic.ttf-4274bf704d3d6b9cd
 Rendering=null
 antialiasing=1
 generate_mipmaps=false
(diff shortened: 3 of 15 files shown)


dbnicholson:
I think this is reasonable, particularly since the flatpak [has been updated to 4.3](https://github.com/flathub/org.godotengine.Godot/commit/bcbdd082883c51eeb75c7ae3fbf7f80006cf34bf), and that's the primary way most of our developers use Godot.

Did you open our project in 4.3? When I've done this before, it's reimported assets. Can you update the tests [to use 4.3](https://github.com/endlessm/godot-block-coding/blob/main/.github/workflows/checks.yaml#L35)? Actually, we should probably run those with a matrix of several older Godot versions since we can't expect users to be update Godot.

dylanmccall:
> I think this is reasonable, particularly since the flatpak [has been updated to 4.3](https://github.com/flathub/org.godotengine.Godot/commit/bcbdd082883c51eeb75c7ae3fbf7f80006cf34bf), and that's the primary way most of our developers use Godot.
> 
> Did you open our project in 4.3? When I've done this before, it's reimported assets. Can you update the tests [to use 4.3](https://github.com/endlessm/godot-block-coding/blob/main/.github/workflows/checks.yaml#L35)? Actually, we should probably run those with a matrix of several older Godot versions since we can't expect users to be update Godot.

Yep, it updates a bunch of .ttf.import files in GUT, adding `disable_embedded_bitmaps=true`. That's from https://github.com/godotengine/godot/commit/911fa38d3a0dc71aa99ffcef474b5bae15805e8d.

I pushed a commit just now with the new import files.

I also checked if there's a newer version of GUT, and there _is_ [a newer version](https://github.com/bitwes/Gut/releases/tag/v9.3.0) but I don't think it has anything relevant to 4.3.

dylanmccall (changes requested):
And I also went and added a matrix of Godot versions for running tests. The setup-godot hook is oddly persnickety about the Godot version ending with a `.0` even though it doesn't actually anywhere else, but all's working :) There is in fact a failing test, which I can confirm fails on my end as well, so, yay, testing! We should fix that.

```
==============================================
= Run Summary
==============================================

res://tests/test_instruction_tree.gd
- test_multiple_entry_script
    [Failed]:  ["extends Node2D
    
    
    func _ready():
    	print({text})
    
    "] expected to equal ["extends Node2D
    
    
    func _ready():
    	print({text})
    
    	print({text})
    
    "]:  
          at line -1

---- Totals ----
Scripts           2
Tests             13
  Passing         12
  Failing         1
Asserts           520
Time              0.176s


---- 1 failing tests ----
```

manuq:
Yes, I see the test failing in my end as well, hmm.

starnight:
I pushed the commit ("[Generate real nodes in test_multiple_entry_script()](https://github.com/endlessm/godot-block-coding/pull/197/commits/6c14f0d5372fd86736e630f2c3a131543cf58f8d)") to fix the error https://github.com/endlessm/godot-block-coding/pull/197#pullrequestreview-2241575407.

manuq:
> I pushed the commit ("[Generate real nodes in test_multiple_entry_script()](https://github.com/endlessm/godot-block-coding/pull/197/commits/6c14f0d5372fd86736e630f2c3a131543cf58f8d)") to fix the error [#197 (review)](https://github.com/endlessm/godot-block-coding/pull/197#pullrequestreview-2241575407).

Excellent, thanks!

manuq:
I tried the different [DuplicateFlags](https://docs.godotengine.org/en/stable/classes/class_node.html#enum-node-duplicateflags) combinations and couldn't make this work. I wonder if we found a regression in 4.3 in the `duplicate()` behavior. In any case, this the code generation is going to change soon, so let's go with @starnight commit.

PR #205, opened 2024-08-23
WIP: Support dragging and dropping block resources to the canvas

dylanmccall:
Ideally, this work will allow us to replace our own DragManager and instead use the drag and drop system built in to Godot.

### DIFF
--- a/addons/block_code/ui/block_canvas/block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/block_canvas.gd
@@ -2,6 +2,7 @@
 extends MarginContainer
 
 const BlockCodePlugin = preload("res://addons/block_code/block_code_plugin.gd")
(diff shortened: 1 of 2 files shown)


manuq:
@dylanmccall is this still useful? I can't imagine a user case for navigating the filesystem dock in search for block resources, then dragging them to the block canvas.

Dragging other things into the canvas to obtain blocks are certainly useful, like the recent drag of a node from the Scene Tree.

dylanmccall:
Yeah, we can close this. I think an interesting use case here is if you have created block resources in your own project and you want to use them, but ideally we'd just teach our code that loads all the block definitions to look for them in other places (instead of in the one directory), so it's kind of suspect anyway.

dylanmccall:
Oh, I'll keep this branch around for a bit - I'd like to do that `DragManager` refactor at some point and there's some relevant stuff here for that.

PR #206, opened 2024-08-29
Add dynamic options lists for blocks

dylanmccall:
We need a way to customize blocks depending on the node being edited. To address this, I found myself doing a few things:

- Replaced the big, global `CategoryFactory` with a `BlockBuilder` that is instantiated for the currently selected block script. I don't super like how I did this - in particular that both picker.gd and block_canvas.gd instantiate their own (redundant) `BlockBuilder` instances. Getting it right means further adjusting how data flows between these things.
- (I don't actually like the name `BlockBuilder` since it implies we're using [the Builder pattern](https://en.wikipedia.org/wiki/Builder_pattern), which was my original intent but it is not what it is doing. Probably just `BlockFactory` would make sense).
- But the big thing this gives us is `BlockBuilder` has a single function which generates a single big list of blocks, including variable blocks. It knows what node our block script belongs to. I think that is a reasonable thing for it to know, but we could specify some rules here like "don't do anything that makes it impossible to move a block script to another node."
- With that said, I broke the variables category here - categories in general are a _disaster_ in this branch - as well as 
`setup_custom_blocks()` in node scripts.
- I added a concept of extension scripts for block definitions. This is an extra GDScript property, so you can optionally attach a plain old Object with some functions defined. I did this for `animationplayer_play.tres`. We should take the time to actually specify an API here, and perhaps a `BlockExtensionScript` base class just to be sure. For now, I just have `animationplayer_play.gd` with `func get_defaults_for_node(context_node: Node)`. It's a little broken still.
- I broke a bunch of tests :(

https://phabricator.endlessm.com/T35564

### DIFF
--- a/addons/block_code/blocks/graphics/animationplayer_play.gd
+++ b/addons/block_code/blocks/graphics/animationplayer_play.gd
@@ -1,0 +1,15 @@
+@tool
+extends BlockExtension
+
+const OptionData = preload("res://addons/block_code/code_generation/option_data.gd")
+
+
+func get_defaults_for_node(context_node: Node) -> Dictionary:
+	var animation_player = context_node as AnimationPlayer
+
+	if not animation_player:
+		return {}
+
+	var animation_list = animation_player.get_animation_list()
+
+	return {"animation": OptionData.new(animation_list)}
--- a/addons/block_code/code_generation/block_ast.gd
+++ b/addons/block_code/code_generation/block_ast.gd
@@ -55,11 +55,15 @@ class ASTNode:
 			# Use parentheses to be safe
 			var argument = arguments[arg_name]
 			var code_string: String
+			var raw_string: String
 			if argument is ASTValueNode:
 				code_string = argument.get_code()
+				raw_string = code_string
 			else:
 				code_string = BlockAST.raw_input_to_code_string(argument)
+				raw_string = str(argument)
 
+			code_block = code_block.replace("{{%s}}" % arg_name, raw_string)
 			code_block = code_block.replace("{%s}" % arg_name, code_string)
 
 		return IDHandler.make_unique(code_block)
@@ -99,11 +103,15 @@ class ASTValueNode:
 			# Use parentheses to be safe
 			var argument = arguments[arg_name]
 			var code_string: String
+			var raw_string: String
 			if argument is ASTValueNode:
 				code_string = argument.get_code()
+				raw_string = code_string
 			else:
 				code_string = BlockAST.raw_input_to_code_string(argument)
+				raw_string = str(argument)
 
+			code = code.replace("{{%s}}" % arg_name, raw_string)
 			code = code.replace("{%s}" % arg_name, code_string)
 
 		return IDHandler.make_unique("(%s)" % code)
@@ -130,15 +138,11 @@ func to_string_recursive(node: ASTNode, depth: int) -> String:
 static func raw_input_to_code_string(input) -> String:
 	match typeof(input):
 		TYPE_STRING:
-			return "'%s'" % input.replace("\\", "\\\\").replace("'", "\\'")
+			return "'%s'" % input.c_escape()
 		TYPE_VECTOR2:
 			return "Vector2%s" % str(input)
 		TYPE_COLOR:
 			return "Color%s" % str(input)
-		TYPE_OBJECT:
-			if input is OptionData:
-				var option_data := input as OptionData
-				return option_data.items[option_data.selected]
 		_:
 			return "%s" % input
 
--- a/addons/block_code/code_generation/block_ast.gd
+++ b/addons/block_code/code_generation/block_ast.gd
@@ -44,35 +44,15 @@ class ASTNode:
 		children = []
 		arguments = {}
 
-	func get_code_block() -> String:
-		var code_block: String = data.code_template  # get multiline code_template from block definition
-
-		# insert args
-
-		# check if args match an overload in the resource
-
-		for arg_name in arguments:
-			# Use parentheses to be safe
-			var argument = arguments[arg_name]
-			var code_string: String
-			var raw_string: String
-			if argument is ASTValueNode:
-				code_string = argument.get_code()
-				raw_string = code_string
-			else:
-				code_string = BlockAST.raw_input_to_code_string(argument)
-				raw_string = str(argument)
-
-			code_block = code_block.replace("{{%s}}" % arg_name, raw_string)
-			code_block = code_block.replace("{%s}" % arg_name, code_string)
-
+	func _get_code_block() -> String:
+		var code_block: String = BlockAST.format_code_template(data.code_template, arguments)
 		return IDHandler.make_unique(code_block)
 
 	func get_code(depth: int) -> String:
 		var code: String = ""
 
 		# append code block
-		var code_block := get_code_block()
+		var code_block := _get_code_block()
 		code_block = code_block.indent("\t".repeat(depth))
 
 		code += code_block + "\n"
@@ -95,25 +75,7 @@ class ASTValueNode:
 		arguments = {}
 
 	func get_code() -> String:
-		var code: String = data.code_template  # get code_template from block definition
-
-		# check if args match an overload in the resource
-
-		for arg_name in arguments:
-			# Use parentheses to be safe
-			var argument = arguments[arg_name]
-			var code_string: String
-			var raw_string: String
-			if argument is ASTValueNode:
-				code_string = argument.get_code()
-				raw_string = code_string
-			else:
-				code_string = BlockAST.raw_input_to_code_string(argument)
-				raw_string = str(argument)
-
-			code = code.replace("{{%s}}" % arg_name, raw_string)
-			code = code.replace("{%s}" % arg_name, code_string)
-
+		var code: String = BlockAST.format_code_template(data.code_template, arguments)
 		return IDHandler.make_unique("(%s)" % code)
 
 
@@ -133,6 +95,26 @@ func to_string_recursive(node: ASTNode, depth: int) -> String:
 		string += to_string_recursive(c, depth + 1)
 
 	return string
+
+
+static func format_code_template(code_template: String, arguments: Dictionary) -> String:
+	for argument_name in arguments:
+		# Use parentheses to be safe
+		var argument_value: Variant = arguments[argument_name]
+		var code_string: String
+		var raw_string: String
+
+		if argument_value is ASTValueNode:
+			code_string = argument_value.get_code()
+			raw_string = code_string
+		else:
+			code_string = BlockAST.raw_input_to_code_string(argument_value)
+			raw_string = str(argument_value)
+
+		code_template = code_template.replace("{{%s}}" % argument_name, raw_string)
+		code_template = code_template.replace("{%s}" % argument_name, code_string)
+
+	return code_template
 
 
 static func raw_input_to_code_string(input) -> String:
--- a/addons/block_code/code_generation/block_ast.gd
+++ b/addons/block_code/code_generation/block_ast.gd
@@ -104,6 +104,14 @@ static func format_code_template(code_template: String, arguments: Dictionary) -
 		var code_string: String
 		var raw_string: String
 
+		if argument_value is OptionData:
+			# Temporary hack: previously, the value was stored as an OptionData
+			# object with a list of items and a "selected" property. If we are
+			# using an older block script where that is the case, convert the
+			# value to the value of its selected item.
+			# See also, ParameterInput._update_option_input.
+			argument_value = argument_value.items[argument_value.selected]
+
 		if argument_value is ASTValueNode:
 			code_string = argument_value.get_code()
 			raw_string = code_string
--- a/addons/block_code/code_generation/block_extension.gd
+++ b/addons/block_code/code_generation/block_extension.gd
@@ -1,3 +1,7 @@
 @tool
 class_name BlockExtension
 extends Object
+
+
+func get_defaults_for_node(context_node: Node) -> Dictionary:
+	return {}
--- a/addons/block_code/code_generation/block_extension.gd
+++ b/addons/block_code/code_generation/block_extension.gd
@@ -1,0 +1,3 @@
+@tool
+class_name BlockExtension
+extends Object
--- a/addons/block_code/serialization/block_script_serialization.gd
+++ b/addons/block_code/serialization/block_script_serialization.gd
@@ -135,7 +135,6 @@ func _get_parameter_block_definition(block_name: String, parameter_name: String)
 
 func _update_block_definitions():
 	_available_blocks.clear()
-	_available_blocks.append_array(_get_default_block_definitions())
 	_available_blocks.append_array(_get_inherited_block_definitions())
 	_available_blocks.append_array(_get_variable_block_definitions())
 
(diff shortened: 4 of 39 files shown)


addons/block_code/code_generation/block_extension.gd
```diff
@@ -0,0 +1,10 @@
+@tool
+
+# FIXME: We should use some kind of prefix for these things. Is this okay?
+class_name EBC_BlockExtension
```
manuq, line 4: Yes I love the idea of a EBC_ prefix for every class that needs a name. We started to follow this pattern in f3cbacb3 for internal scripts:

```
const BlockExtension = preload("res://addons/block_code/code_generation/block_extension.gd")
```

But this doesn't allow extending the class like in `extends EBC_BlockExtension`.

dylanmccall:
Okay, finally back on track here. The remaining work is pretty much go through and delete some now-obsolete code, fix some GUT tests, and fix one stubborn new bug (probably from clumsy rebasing) where a block script isn't loading if it has loose non-Entry blocks placed in it.

While I was here, I tore out some unnecessary optimization around `BlocksCatalog` / `BlockFactory` / `CategoryFactory`. I feel a bit guilty doing it, but I don't think we need to do any kind of indexed lookups here. A function that iterates through all the blocks and looks at their category names and then forgets about it is fine for the scale at hand.

A brave person in the future might come back and improve our plumbing a bit so we're passing a single `BlockFactory` from _main_panel.gd_ down to its children instead of creating two different ones (one for the canvas and one for the picekr). But I think we're fine for the moment.

There's a bunch here, but one new thing to highlight: I didn't like how `BlocksCatalog` was [generating a bunch of block definitions ahead of time](https://github.com/endlessm/godot-block-coding/blob/3b78183394f29ea4b6e9e6879ef7f389a9ad412c/addons/block_code/code_generation/blocks_catalog.gd#L133-L197) for output parameters. So I took that out, and now `BlockFactory` has a single `instantiate_block_by_name` which has some smarts to generate a block right there if the name follows a particular "block_name:parameter" format.

To make that work, I also changed _block_script_serialization.gd_ so all of our AST <-> block conversions are storing just the block name, instead of the whole block definition. (This was necessary because it doesn't know enough to deal with block factories, but previously it was talking to `BlocksCatalog`). I _think_ that's fine (it means later on we instantiate from the name instead of from the definition), but if you know confidently either way I'd love to know before I break everything with that change :) If it does break everything, I have a slightly more complicated alternative in mind.

dylanmccall:
One last opinionated change here: my template editor / options list stuff came with a side-effect in our generated code: the AST to code block stuff helpfully wraps strings in quote marks, except when those strings come from an options list. We can't do that anymore since the value it sees is always just a string now :)

To address that, I added a way to specify a parameter substitution that _won't_ be quoted: just `{{parameter}}` instead of `{parameter}`.

addons/block_code/blocks/graphics/animationplayer_play.gd
```diff
@@ -0,0 +1,15 @@
...
+extends BlockExtension
+
+const OptionData = preload("res://addons/block_code/code_generation/option_data.gd")
+
+
+func get_defaults_for_node(defaults: Dictionary, context_node: Node) -> Dictionary:
+	var animation_player = context_node as AnimationPlayer
+
+	if not animation_player:
+		return {}
```
dbnicholson, line 11: Shouldn't this return `defaults`?

dbnicholson (commented):
I haven't gotten too far, but I wanted to share some review so far since I have to run around a bit today.

manuq (changes requested):
This is working like a charm for animations! I will leave to @dbnicholson the code review. I appreciate how the last commit improves the UI for the "is action pressed/released" block, which was extra large. On the other hand, I see a bug in that one putting the action name verbatim when it should be quoted for it to work:
Captura desde 2024-09-13 16-31-14
This then fails with error:
```
Parser Error: Identifier "ui_up" not declared in the current scope.
```

addons/block_code/code_generation/block_extension.gd
dbnicholson: Is that... a bike shed?! I feel like we either want underscores or camel case but not both. So, `EbcSomeClass` seems like the natural continuation of the camel case way.

addons/block_code/ui/blocks/utilities/template_editor/template_editor.gd
```diff
@@ -0,0 +1,137 @@
...
+
+	if default_value is OptionData:
+		var option_data := default_value as OptionData
+		parameter_input.option_data = option_data
+		if option_data.selected < option_data.items.size():
+			parameter_input.default_value = option_data.items[option_data.selected]
+	else:
+		parameter_input.default_value = default_value
+
+	parameter_input.modified.connect(modified.emit)
```
dbnicholson, line 124: It looks like this is causing an error on 4.2 import:

```
SCRIPT ERROR: Parse Error: Cannot find property "emit" on base "Signal".
          at: GDScript::reload (res://addons/block_code/ui/blocks/utilities/template_editor/template_editor.gd:124)
SCRIPT ERROR: Parse Error: Cannot find member "emit" in base "Signal".
          at: GDScript::reload (res://addons/block_code/ui/blocks/utilities/template_editor/template_editor.gd:124)
```
It looks like the way we've been handling this elsewhere is to make a lambda:

```suggestion
	parameter_input.modified.connect(func(): modified.emit())
```

dbnicholson (changes requested):
I kinda glazed over on TemplateEditor and the handling of options, but I trust your handling there. My two main things I'd want to see:

1. Don't make `BlockScriptSerialization` dependent on `BlockFactory` since it's "recoupling" the UI and the data. It's fine if you want to shuffle around some parts, but I don't see any strong reasons why `BlockFactory` shouldn't be a thin `BlockDefinition` to `Block` layer on top of `BlockScriptSerialization`.
2. The bug @manuq found should be resolved.

addons/block_code/ui/block_factory.gd
```diff
@@ -0,0 +1,236 @@
...
+	Types.BlockType.STATEMENT: preload("res://addons/block_code/ui/blocks/statement_block/statement_block.tscn"),
+	Types.BlockType.VALUE: preload("res://addons/block_code/ui/blocks/parameter_block/parameter_block.tscn"),
+	Types.BlockType.CONTROL: preload("res://addons/block_code/ui/blocks/control_block/control_block.tscn"),
+}
+
+var block_script: BlockScriptSerialization
+var block_code_node: BlockCode
+var context_node: Node:
+	get:
+		return block_code_node.get_parent() if block_code_node is Node else null
```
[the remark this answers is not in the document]
dylanmccall: In an earlier version of this I had BlockFactory setting defaults from an extension script itself, but I ended up moving that to Block itself so this is indeed very superfluous here. It _might_ need to come back later if we want our set of available blocks to involve other nodes in the scene, which I had in mind here but it isn't a thing we've reached any decisions on.

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -71,46 +44,46 @@ func get_categories() -> Array[BlockCategory]:
 	return []
 
 
-func generate_ast_list() -> ASTList:
+func generate_ast_list(block_factory: BlockFactory) -> ASTList:
```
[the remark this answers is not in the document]
dylanmccall: Oops! The circular dependency will be resolved with the other change removing `BlockFactory.block_node`, and I'll see if that lands us on `block_factory` being a property of BlockScript (which would certainly make some sense).

addons/block_code/ui/block_factory.gd
```diff
@@ -0,0 +1,236 @@
...
+			script.setup_custom_blocks()
+
+	update_block_definitions()
+
+
+func update_block_definitions():
+	_available_blocks.clear()
+	_available_blocks.append_array(_get_default_block_definitions())
+	_available_blocks.append_array(_get_inherited_block_definitions())
+	_available_blocks.append_array(_get_variable_block_definitions())
```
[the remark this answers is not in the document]
dylanmccall: I'll shuffle this around a bit and see if I can flatten these things out a bit and it should reduce the diff a bunch. Mainly this happened because I was frustrated with the number of different places where we gather up block definitions, although that frustration was more with the code before #147. Looking at this at the moment, I'm extremely annoyed that `_get_variable_block_definitions` is accessing `block_script.variables`, so _maaaybe_ BlockScript should be more involved there. I'll leave a note once when I've addressed that!

addons/block_code/ui/block_factory.gd
```diff
@@ -0,0 +1,236 @@
...
+
+
+func _init(p_block_script: BlockScriptSerialization = null, p_block_code_node: Node = null):
+	block_script = p_block_script
+	block_code_node = p_block_code_node
+
+	_init_block_definitions()
+
+
+func instantiate_block(block_definition: BlockDefinition) -> Block:
```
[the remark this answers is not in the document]
dylanmccall: It ended up like that so we can get at the list of available categories and set `Block.color` accordingly. This way we can lose `Util.get_category_color`, and for instance, DragManager doesn't need to copy the color property in its `copy_block` function. (Although actually I forgot to remove that).

But really the only reason for `Block.block_factory` is so the ParameterOutput component has something on which to call `_block_factory.instantiate_block_by_name()`. That whole scheme should be improved, but I think we'd want to dig in to DragManager to do it. For now, perhaps I can make ParameterOutput and DragManager get this information in a more special way (the current BlockCanvas could just be a static somewhere) so we don't need to have a general-purpose `Block.block_factory`.

addons/block_code/blocks/graphics/animationplayer_play.gd
dylanmccall: Thanks!

But I have a question while we're here: do you feel it's better if this function returns a complete defaults list (eg: returning `{}` means whatever defaults were there are now _gone_), or should it be returning just new defaults (so the caller is doing `defaults.merged(defaults, new_defaults)` itself). It doesn't make a difference either way in this case so I just went for the thing that makes the least assumptions.

dylanmccall:
> This then fails with error:
> 
> ```
> Parser Error: Identifier "ui_up" not declared in the current scope.
> ```

Oops! That happened because I _removed_ some quote marks that were supposed to be there while I was tweaking all the blocks with options lists. The action_name property is always a STRING_NAME so it needed to be quoted. I fixed it by changing the code template to say `"Input.is_action_{{action}}('{{action_name}}')"` (with the new `{{param}}` syntax) so it's clear what we're expecting). And I changed the display template to have the correct type for that parameter, but option lists are troublesome because we just end up with whatever type is in the options list regardless of the parameter type. Maybe something should print a warning when there's a mismatch :thinking:

addons/block_code/blocks/graphics/animationplayer_play.gd
dbnicholson: I feel like it should merge until we have evidence that it's better to override defaults from a base class.

addons/block_code/ui/block_factory.gd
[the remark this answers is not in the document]
dbnicholson: Maybe the thing to do here is to keep a reference to either the current `BlockScriptSerialization` or `BlockCodeNode` instance. One or the other encompasses the block coding scene with which you'd have custom or property specific block definitions. With access to the current `BlockScriptSerialization`, you can query or add `BlockDefinition`s as needed.

`instantiate_block_by_name` is really nothing more than finding the `BlockDefinition` and creating a `Block` scene from it. Similarly, the available categories are already stored in the `BlockScriptSerialization` instance and getting the color is just a straight dictionary lookup.

dylanmccall:
Thank you for your thorough review @dbnicholson ! It was super helpful. I got rid of the BlockFactory stuff, merged things back _somewhat_, but pulled that not-quite-one-time BlockCatalog setup outside of BlockScript and added a `BlockEditorContext` singleton which gives us effectively what I was getting out of that previous arrangement: a single thing that says "this is the block code node and block script we are looking at," along with a signal which fires when that changes.

Need to squash some commits and apparently fix some lingering issues with Godot 4.2.2 :cry: - but those should be small things. How's it looking at this point?

addons/block_code/blocks/graphics/animationplayer_play.gd
dylanmccall: Also it turns out Dictionary.merged is new in Godot 4.3, so better to do the ugly alternative in one place anyway.

dylanmccall:
This is all rebased, passing tests, and ready to review :)

addons/block_code/ui/blocks/utilities/template_editor/template_editor.gd
```diff
@@ -114,13 +114,11 @@ func _append_input_parameter(parameter: Dictionary, id: int):
 	parameter_input.placeholder = parameter["name"]
 	parameter_input.variant_type = parameter["type"]
 
-	if parameter["is_option"] and default_value is OptionData:
```
dbnicholson: FWIW, `is_option` is never set in the current PR. I guess it must have existed in an earlier version.

addons/block_code/blocks/graphics/animationplayer_play.gd
dbnicholson: OK, I see now that this returns additional defaults and BlockDefinition does the merging.

dbnicholson (approved):
This is great! I left some nitpicks and comments, but nothing that would stop this from being merged now. I'm going to leave this for a couple hours to see if @manuq has any further input, and then I'll merge.

manuq:
@dylanmccall @dbnicholson I gave it a quick test, and found the following error. Adding a "change rotation by" to the "on process" leaves the `_on_process` empty, thus failing with syntax error.
Captura desde 2024-09-19 15-09-34

addons/block_code/ui/block_editor_context.gd
```diff
@@ -0,0 +1,38 @@
+class_name BlockEditorContext
```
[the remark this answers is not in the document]
manuq: I second that!

addons/block_code/ui/blocks/utilities/template_editor/template_editor.gd
```diff
@@ -0,0 +1,132 @@
+@tool
+class_name TemplateEditor
```
manuq, line 2: I really like the TemplateEditor, much better than the formatting used before.

addons/block_code/ui/block_editor_context.gd
```diff
@@ -0,0 +1,38 @@
...
+		return block_code_node.block_script
+
+var parent_node: Node:
+	get:
+		if block_code_node == null:
+			return null
+		return block_code_node.get_parent()
+
+
+func set_block_code_node(value: BlockCode):
```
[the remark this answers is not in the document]
dylanmccall: Yeah, I'll do that :) I had it as a setter at one point and then it was doing other stuff so I felt better with it being a function, but it isn't doing other stuff at the moment.

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -176,64 +175,6 @@ func load_object_script() -> Object:
 	return null
 
 
-func _get_default_block_definitions() -> Array[BlockDefinition]:
```
manuq: Great code removal! Having this dynamic is much better.

dylanmccall:
> @dylanmccall @dbnicholson I gave it a quick test, and found the following error. Adding a "change rotation by" to the "on process" leaves the `_on_process` empty, thus failing with syntax error. Captura desde 2024-09-19 15-09-34
> 
> EDIT
> 
> In the output I see:
> 
> ```
>   res://addons/block_code/code_generation/block_ast.gd:120 - Trying to assign value of type 'float' to a variable of type 'String'.
>   res://addons/block_code/code_generation/block_ast.gd:120 - Trying to assign value of type 'float' to a variable of type 'String'.
> ```

Oops, that was a dumb mistake in 0bd3c26189adc755faa7bcc0d5973ef628879115. Fixed :)

addons/block_code/code_generation/block_ast.gd
```diff
@@ -104,6 +104,14 @@ static func format_code_template(code_template: String, arguments: Dictionary) -
 		var code_string: String
 		var raw_string: String
 
+		if argument_value is OptionData:
+			# Temporary hack: previously, the value was stored as an OptionData
+			# object with a list of items and a "selected" property. If we are
+			# using an older block script where that is the case, convert the
+			# value to the value of its selected item.
+			# See also, ParameterInput._update_option_input.
+			argument_value = argument_value.items[argument_value.selected]
```
manuq, line 113: I don't think that we need to maintain backwards compatibility, there has been wild refactors since the last release. It doesn't harm to have this anyways.

manuq (approved):
All looks good from my side and the last bug I found is already fixed.

addons/block_code/ui/blocks/control_block/control_block.tscn
```diff
@@ -95,4 +95,5 @@ shift_top = 20.0
 [node name="SnapPoint" parent="VBoxContainer" instance=ExtResource("3_nhryi")]
 layout_mode = 2
 
+[connection signal="modified" from="." to="." method="_modified"]
```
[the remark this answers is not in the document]
dylanmccall: Oh, yes, that was connected to the wrong thing! Ugh, that's a dreadful category of errors. I just remembered block.gd actually knows about its TemplateEditor now, so it can do all of the signal propagation in gdscript instead of the current scheme.

PR #207, opened 2024-08-29
Check _selected_block_code is still valid before disconnecting

dylanmccall:
Previously, it was possible for _selected_block_code to refer to an object which had been freed, which resulted in an error when calling disconnect for its signals.


starnight (approved):
LGTM!

PR #210, opened 2024-08-30
Fix Area2d blocks

manuq:
Fix regression while moving blocks to separate resource files.


dylanmccall (changes requested):
I would like it if the commit message described what the regression was. It's unclear from looking at the changes, and I'm surprised we'd prefer `area_entered` (just `Area2D` nodes) over `body_entered` (`PhysicsBody2D` nodes, like `CharacterBody2D` and `RigidBody2D`) based on how they're described in the Godot docs.

manuq:
> I would like it if the commit message described what the regression was. It's unclear from looking at the changes, and I'm surprised we'd prefer `area_entered` (just `Area2D` nodes) over `body_entered` (`PhysicsBody2D` nodes, like `CharacterBody2D` and `RigidBody2D`) based on how they're described in the Godot docs.

Oh I thought it was a regression but no. I did the port right in 4a1e52a7. It's the names that are misleading "area2d_on_entered/exited". I see that Area2D has both signals.

In my demo games I was expecting to do things by detecting collisions of shapes, without a rigid body. So I should propose this as new blocks, not replacing the current ones.

PR #211, opened 2024-09-03
examples: Pachinko

cassidyjames:
WIP physics-based game built entirely with blocks!

Screenshot from 2024-09-03 14-35-39

Choose where to drop the ball, then it bounces on the pins until it lands in a prize/goal receptacle. I'm also thinking it would be neat to allow a few "bumps" (could count as using lives) to influence how the ball drops.

This uses the built-in physics to do a lot of the work, but also uses the new "Random" blocks for the bumps (it changes the velocity by a random amount) and can demonstrate how to use "lives" in other interesting ways.

- [ ] Am I doing the "Pins" in an overly-complicated way (repeating them manually)? I feel like I'm missing something…
- [ ] Add main prize/goal
- [ ] Add bump counter
- [ ] Sound effects

https://phabricator.endlessm.com/T35633


cassidyjames:
Moving to https://github.com/endlessm/godot-block-coding-demos

PR #214, opened 2024-09-05
Remove _exit_tree methods from Simple nodes

### LINKED ISSUE #213: SimpleCharacter texture can dissapear while editing
### What Happened

If I switch to another scene, the texture dissapears from the editor:
[Grabación de pantalla desde 2024-09-05 10-14-19.webm](https://github.com/user-attachments/assets/c0c705bf-7a53-4c14-8d44-97a1b22adfe3)

Trying to set the texture again shows an error in the Output panel (see below).

By the way in the game it looks fine.

### How to Reproduce

1. Add a SimpleCharacter node
2. Add a texture to it
3. Switch to another scene, then back

### Block Coding Plugin Version

main

### Godot Engine Version

v4.3

### Operating System

Endless OS

### Logs or command-line output

```shell
res://addons/block_code/simple_nodes/simple_character/simple_character.gd:47 - Invalid assignment of property or key 'texture' with value of type 'CompressedTexture2D' on a base object of type 'Nil'.
```

### Anything else?

_No response_

manuq:
There is no need for this cleanup because from docs: [1]
> If the node has children, its _exit_tree callback will be called last,
> after all its children have left the tree.

Also because: [2]
> when a node is freed with Object.free or queue_free, it will also free
> all its children.

Further, this removes the sprite and collision while in the editor when the user switches to another scene tab #213.

[1] https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-private-method-exit-tree
[2] https://docs.godotengine.org/en/stable/classes/class_node.html#description

Fixes #213

### DIFF
--- a/addons/block_code/simple_nodes/simple_character/simple_character.gd
+++ b/addons/block_code/simple_nodes/simple_character/simple_character.gd
@@ -72,16 +72,6 @@ func simple_setup():
 	_texture_updated()
 
 
-func _exit_tree():
-	if collision:
-		collision.queue_free()
-		collision = null
-
-	if sprite:
-		sprite.queue_free()
-		sprite = null
-
-
 func get_custom_class():
 	return "SimpleCharacter"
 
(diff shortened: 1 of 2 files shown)


dbnicholson (approved):
Good call. I totally misunderstood what `_exit_tree` was for. I think the reason I added that was that I was worried that `sprite` and/or `collision` would be created but then changed before before they were added to the node. But that can't happen with the current code since their created and added all within `_ready`.

PR #217, opened 2024-09-10
Add character states

manuq:
Add ways to check different states of the character, so a platformer character can have different assets / animations when jumping / moving / being idle.

Example video:
[Grabación de pantalla desde 2024-09-10 15-04-11.webm](https://github.com/user-attachments/assets/e6e235e9-9c6f-4666-b677-be66f5a9aea7)

That is using these blocks:
Captura desde 2024-09-10 15-05-01


starnight (approved):
Thanks!  LGTM!

PR #221, opened 2024-09-13
Improve keyboard accessibility in create_variable_dialog

dylanmccall:
When the dialog is focused, we need to focus the custom dialog's own controls instead of the dialog's buttons. In addition, set focus_next and focus_previous properties so focus flows between child widgets and the dialog's ok / cancel buttons. The focus loops around within the dialog, as is conventional with other dialogs in the Godot editor.


starnight (approved):
Awesome!  Thanks!

PR #222, opened 2024-09-18
Add is_input_actioned.tres

dylanmccall:
This adds a block extension script to generate an options list instead of `BlocksCatalog._get_inputmap_actions`.

### DIFF
--- a/addons/block_code/blocks/input/is_input_actioned.gd
+++ b/addons/block_code/blocks/input/is_input_actioned.gd
@@ -1,0 +1,35 @@
+@tool
+extends BlockExtension
+
+const OptionData = preload("res://addons/block_code/code_generation/option_data.gd")
+
+
+func get_defaults_for_node(_context_node: Node) -> Dictionary:
+	var inputmap_actions = _get_inputmap_actions()
+	return {"action_name": OptionData.new(inputmap_actions)}
+
+
+static func _get_inputmap_actions() -> Array[StringName]:
+	var inputmap_actions: Array[StringName]
+
+	var editor_input_actions: Dictionary = {}
+	var editor_input_action_deadzones: Dictionary = {}
+	if Engine.is_editor_hint():
+		var actions := InputMap.get_actions()
+		for action in actions:
+			if action.begins_with("spatial_editor"):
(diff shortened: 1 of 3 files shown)


addons/block_code/blocks/input/is_input_actioned.gd
```diff
@@ -0,0 +1,35 @@
+@tool
+extends BlockExtension
+
+const OptionData = preload("res://addons/block_code/code_generation/option_data.gd")
```
starnight, line 4: **const OptionData** has been moved to here.  Should the [original one](https://github.com/endlessm/godot-block-coding/pull/222/files#diff-ec1a3a88731ab0f21c32be474930ad0bceb7078d57c246d16a3c50ebd9dffadeL4) in "addons/block_code/code_generation/blocks_catalog.gd" be dropped?

addons/block_code/blocks/input/is_input_actioned.tres
```diff
@@ -0,0 +1,27 @@
...
+[sub_resource type="Resource" id="Resource_ai5in"]
+script = ExtResource("1_d8i05")
+selected = 0
+items = ["pressed", "just_pressed", "just_released"]
+
+[resource]
+script = ExtResource("1_rprh7")
+name = &"is_input_actioned"
+target_node_class = ""
+description = "True if the character is on the floor."
```
starnight, line 16: How about modifying the description as "Return the input key action"?

starnight (changes requested):
Thanks for helping move the input key action block into a single tres.  I place some review comments. :)

addons/block_code/blocks/input/is_input_actioned.gd
dylanmccall: :+1:

addons/block_code/blocks/input/is_input_actioned.tres
dylanmccall: Thanks for noticing! That's fixed in the newest commit.

dylanmccall:
Thanks for the review @starnight! I fixed those details in the newest commit.

starnight (approved):

PR #224, opened 2024-09-19
Add breakpoint block

dylanmccall:
An example of this in use:

Screenshot From 2024-09-19 12-02-53

We can't see the generated code in the debugger (or at least I don't know how to reach it), but we can inspect local variables and the Godot objects they are referencing, which can be fairly useful.


starnight (approved):
Interesting feature!  Thanks!

I add a **Breakpoint** block into BlockCode directly:
image
The generated code is:
```gdscript
extends SimpleCharacter


func _process(delta):
    move_with_player_buttons('player_1', 'platformer', delta)
    if (Input.is_action_just_pressed('ui_right')):
        rotation_degrees += 5
        position += Vector2(1, 0)
        breakpoint
    if (Input.is_action_pressed('ui_left')):
        rotation_degrees += -5
        position += Vector2(-1, 0)
```
It breaks at it and shows the Stack Trace: 
image

manuq:
Nice!

PR #227, opened 2024-09-24
Allow dragging nodes to block canvas

dbnicholson:
In order to call methods on other nodes, you need a reference to the node. This adds a new `get_node` block definition that has all the other nodes provided as options. It also adds a method to drag a node from the scene tree dock and set the node's path in a `get_node` block automatically.

Unfortunately, in order to make the drag action usable, I had to drop the multi-select feature that automatically activates the block node's parent in the scene editor. I searched long for a way to do that without selecting the node, but it doesn't appear that there's a way to do it.

https://phabricator.endlessm.com/T35648

### DIFF
--- a/addons/block_code/blocks/communication/get_node.tres
+++ b/addons/block_code/blocks/communication/get_node.tres
@@ -1,0 +1,27 @@
+[gd_resource type="Resource" load_steps=5 format=3 uid="uid://canpdkahokjqs"]
+
+[ext_resource type="Script" path="res://addons/block_code/code_generation/option_data.gd" id="1_bk47y"]
+[ext_resource type="Script" path="res://addons/block_code/code_generation/block_definition.gd" id="1_d60g7"]
+[ext_resource type="Script" path="res://addons/block_code/blocks/communication/get_node.gd" id="1_we5wl"]
+
+[sub_resource type="Resource" id="Resource_esr4a"]
+script = ExtResource("1_bk47y")
+selected = 0
+items = []
+
+[resource]
+script = ExtResource("1_d60g7")
+name = &"get_node"
+target_node_class = ""
+description = "Get the node at the given path"
+category = "Communication | Nodes"
+type = 3
+variant_type = 24
+display_template = "{path: NIL}"
+code_template = "get_node(\"{path}\")"
+defaults = {
+"path": SubResource("Resource_esr4a")
+}
+signal_name = ""
+scope = ""
+extension_script = ExtResource("1_we5wl")
--- a/addons/block_code/ui/block_canvas/block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/block_canvas.gd
@@ -60,6 +60,45 @@ func _ready():
 		_open_scene_button.icon = _open_scene_icon
 
 
+func _can_drop_data(at_position: Vector2, data: Variant) -> bool:
+	if _context.block_code_node == null or _context.parent_node == null:
+		return false
+	if typeof(data) != TYPE_DICTIONARY:
+		return false
+
+	var nodes: Array = data.get("nodes", [])
+	if nodes.size() != 1:
+		return false
+	var abs_path: NodePath = nodes[0]
+
+	# Don't allow dropping BlockCode nodes or nodes that aren't part of the
+	# edited scene.
+	var node := get_tree().root.get_node(abs_path)
+	if node is BlockCode or not Util.node_is_part_of_edited_scene(node):
+		return false
+
+	# Don't allow dropping the BlockCode node's parent as that's already self.
+	var parent_path: NodePath = _context.parent_node.get_path()
+	return abs_path != parent_path
+
+
+func _drop_data(at_position: Vector2, data: Variant) -> void:
+	var abs_path: NodePath = data.get("nodes", []).pop_back()
+	if abs_path == null:
+		return
+
+	# Figure out the best path to the node.
+	var node := get_tree().root.get_node(abs_path)
+	var node_path: NodePath = Util.node_scene_path(node, _context.parent_node)
+	if node_path in [^"", ^"."]:
+		return
+
+	var block = _context.block_script.instantiate_block_by_name(&"get_node")
+	block.set_parameter_values_on_ready({"path": node_path})
+	add_block(block, at_position)
+	reconnect_block.emit(block)
+
+
 func add_block(block: Block, position: Vector2 = Vector2.ZERO) -> void:
 	if block is EntryBlock:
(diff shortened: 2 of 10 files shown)


addons/block_code/blocks/communication/get_node.tres
```diff
@@ -0,0 +1,27 @@
...
+
+[resource]
+script = ExtResource("1_d60g7")
+name = &"get_node"
+target_node_class = ""
+description = "Get the node at the given path"
+category = "Communication | Nodes"
+type = 3
+variant_type = 24
+display_template = "Node {path: NIL}"
```
manuq, line 20: I would either remove the "Node" prefix from here or from the "Call method in node NODE" block, otherwise the word is repeated:
Captura desde 2024-09-24 08-41-05
```suggestion
display_template = "{path: NIL}"
```

manuq (changes requested):
It works like a charm! Thanks @dylanmccall to make this so easy. I have only added a fixup commit to pass the tests, and added a suggestion to remove the repeated "node" word.

addons/block_code/blocks/communication/get_node.tres
manuq: Yeah.. I think in the future we should have a small icon next to the label of value blocks indicating their type (Vector2, float, object, etc). Scratch has pointy blocks for booleans but in Godot there are more types so it will require too many shapes.

dylanmccall:
One note about the multi-select behaviour: it was also so, when someone is editing a block code node, the node it belongs to is highlighted in the canvas. I think it's fine losing this since it didn't work super well anyway and this feature is _far_ more important, but it was something that was casually requested at some point. A better approach would be actually adding a representation of BlockCode itself in the canvas, but that would have to be pretty clever since it would depend on what type the parent node is.

addons/block_code/ui/util.gd
```diff
@@ -12,3 +12,38 @@ static func node_is_part_of_edited_scene(node: Node) -> bool:
...
+## Returns the path from [param reference] to [param node] without referencing
+## parent nodes. If [param node] is [param reference] or a child of it in the
+## scene tree, a relative path is returned. If [param node] is an ancestor of
+## [param reference] in the scene tree, an absolute path using [param
+## path_root] is returned.
+## [br]
+## Both [param node] and [param reference] must be ancestors of [param
+## path_root]. If [param path_root] is [constant null]
+## [method EditorInterface.get_edited_scene_root().get_parent] is used.
+static func node_scene_path(node: Node, reference: Node, path_root: Node = null) -> NodePath:
```
manuq, line 28: This is great. I wonder if you had to read the editor source for reference or implemented it from scratch.

manuq (approved):

manuq:
> One note about the multi-select behaviour: it was also so, when someone is editing a block code node, the node it belongs to is highlighted in the canvas. I think it's fine losing that since it didn't work super well anyway and this feature is _far_ more important, but it was something that was casually requested at some point. A better approach would be actually adding a representation of BlockCode itself in the 2D / 3D canvas, but that would have to be pretty clever since it would depend on what type the parent node is.

Ah yes, that would be much welcome. A way to highlight the target node. I think that can be done with an editor plugin for any node that inherits CanvasItem by using [get_viewport_rect()](https://docs.godotengine.org/es/4.x/classes/class_canvasitem.html#class-canvasitem-method-get-viewport-rect) to get the bounding box.

addons/block_code/ui/util.gd
dbnicholson: It was a lot of trial and error and printing out paths until I felt reasonably comfortable that I knew the pattern. I also had to read the SceneTree, NodePath and Node documentation many times. In particular, I didn't understand that `get_edited_scene_root()` or `SceneTree.edited_scene_root` _don't_ correspond to `/root`. `/root` corresponds to the parent of the scene root node. That's `SceneTree.root`, but in the editor that's the entire editor window. Hence why it does `get_edited_scene_root().get_parent()` since you're deep in the editor hierarchy in our cases.

I'm sure there are some corner cases that aren't covered, but I think it will fit our needs.

manuq: Yeah this is certainly something to improve upstream. Why dragging a node from the current scene tree would give you a path relative to the editor root?

dbnicholson: Because the current scene _is_ the whole editor. The path you're given is an absolute path that's entirely valid while the editor is running because that's where it is in the (singleton) `SceneTree`.

I should have put this in a comment, but this is a special case for us. We're trying to create an absolute path for when `edited_scene_root` later becomes `current_scene`. The absolute '/root` path returned from here is bogus while you're in the editor. It only becomes valid when the scene is actually running. That's what we need here because we're generating a script that won't actually run until the edited scene becomes the current scene and is at the root of the scene tree.

If you just want an absolute path to the node in whatever is the current scene, you can just use `Node.get_path()`. Originally I wrote this to always use relative paths as those work no matter where the scene is in the scene tree. That's also why you use relative paths such as `$Foo/Bar` (equivalent to `get_node("Foo/Bar")`) in GDScript - they work regardless of where the node using the script is in the scene tree. However, relative paths with `../../SomeNode` seemed too hard to grasp, which is why I came up with the  "absolute path that will become valid at runtime" thing.

So, if you wanted to upstream this, it would have to be a separate method like `get_edited_path(relative: bool = false)` that returns a path treating the edited scene as if it was the current scene.

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -60,6 +60,45 @@ func _ready():
...
+	# Figure out the best path to the node.
+	var node := get_tree().root.get_node(abs_path)
+	var node_path: NodePath = Util.node_scene_path(node, _context.parent_node)
+	if node_path in [^"", ^"."]:
+		return
+
+	var block = _context.block_script.instantiate_block_by_name(&"get_node")
+	block.set_parameter_values_on_ready({"path": node_path})
+	add_block(block, at_position)
+	reconnect_block.emit(block)
```
dylanmccall, line 99: I have a suggestion for some future work here. Instead of setting the value of an option parameter in the get_node block, we could use [the same scheme as parameter blocks](https://github.com/endlessm/godot-block-coding/blob/main/addons/block_code/serialization/block_script_serialization.gd#L82-L133), so in this spot we'd have something like `instantiate_block_by_name("get_node:${node_path}")` and it would generate a unique parameter block just for that node path (instead of a block with a big options list and the value set to the node path).

The hard part would be fixing the inevitable small issues which arise from doing that, since right now we only use the `block_name:detail` syntax for the one thing, but this use case is definitely something I had in mind when I added it :)

PR #228, opened 2024-09-24
Block picker enhancements

dylanmccall:
- In the picker, never destroy components. Instead, show and hide category buttons and category display components as needed.
- Simplify the category display component by having it take a list of block definitions as a property instead of the component fetching the list of blocks in a category itself.
- In the category display component, reuse block components instead of always creating new ones. In my testing, this has resulted in a noticeable performance improvement when switching between block code nodes. (At the moment, the list of blocks is never pruned. There isn't actually an upper limit because variable blocks are blocks as well, but I don't see this being a problem. At least not for a long time).
- Tweak the Variables blocks category slightly. We still have variable_category_display.tscn, but I styled the create variable button to be a bit more buttony, and I moved it to the bottom of the list instead of the top. I did this because creating a variable is actually really smooth now! The variable blocks just appear there without a noticeable delay, and without affecting the scroll position. We could animate them if we wanted to and it would be fun. (And this all means we want them to actually be in frame). I also moved the Vector2 block out of the Variables category. I don't know why it was there, but it was the only thing making that category visible, which bugged me, so I fixed that by special-casing BlockScriptSerialization to make the Variables category _always_ visible.
- Change `BlockExtension` to be instantiated per block, instead of per block definition. This makes sense because they deal in UI and they didn't really belong in `BlockDefinition` anyway.
- To go along with that, deprecate `BlockExtension.get_defaults_for_node()` in favour of `BlockExtension.get_defaults()`. The context node property is now set for each instance.
- Improve documentation for the BlockExtension class.
- Add a way for a block extension to notify its block that is has changed. This means we can update specific blocks as needed.
- Change the _animationplayer_play_ block extension to emit the new "changed" event when the user changes animations in the animation player. This means the block will always have an up to date options list.


addons/block_code/ui/blocks/block/block.gd
```diff
@@ -27,7 +27,16 @@ signal modified
 @export var scope: String = ""
 
 ## The resource containing the definition of the block
-@export var definition: BlockDefinition
+@export var definition: BlockDefinition:
+	set(value):
+		var is_changed := definition != value
+		if definition and is_changed:
+			definition.changed.disconnect(_on_definition_changed)
```
manuq, line 34: The amount of times that I wanted to hook to the changed signal of custom resources. Thanks!

addons/block_code/code_generation/block_extension.gd
```diff
@@ -8,13 +8,28 @@
...
+## as [method get_defaults].[br]
+## [br]
+## In some cases, an extension may need to monitor the scene to determine if it
+## has changed. To achieve this, override [method _context_node_changed] to
+## connect the relevant signals, and call [method _emit_changed] when a
+## significant change has occurred.
 class_name BlockExtension
 extends RefCounted
 
+signal changed
```
manuq, line 20: Oh wait I thought by looking at the `definition.changed.connect` in previous commit that you were going to use the `changed` signal already available in Resource https://docs.godotengine.org/en/stable/classes/class_resource.html#class-resource-method-emit-changed . But this is for the extension, got it.

manuq (approved):
This is impressive. I note the performance boost too.

addons/block_code/code_generation/block_extension.gd
dylanmccall: To be honest I thought that too, then I ended up doing that whole thing moving block extensions closer to Block, but then I left it because we probably _should_ connect to that changed signal as well. I really wish changed was in the Object class so it didn't feel weird copying Resource here. Some day maybe.

PR #230, opened 2024-09-25
Add SimpleSpawner node and blocks

manuq:
This is a node with blocks to simplify the instantiation of other scenes and adding them to the current one.

Instantiated scenes are added as children of the SimpleSpawner node, so their position, ordering, etc, is relative to this node.

Blocks are provided to spawn once and to start / stop spawning based on a frequency.

Multiple scenes can be provided to be instantiated. In which case one is randomly picked when spawning.

For performance concerns, the amount of instantiated scenes is limited by default, unless the learner sets the limit to be zero (unlimited).

There are 2 behaviors to choose for when the limit is reached:
- Replace: Remove the oldest spawned scene and spawn a new one.
- No Spawn: No spawn happens until any spawned scene is removed by other means.

https://phabricator.endlessm.com/T35662


starnight (approved):
LGTM!

PR #231, opened 2024-09-25
Switch block labels to natural language

manuq:
Eg. use "Remove" instead of "Queue Free".

Also, remove the "On physics process" block. This is confusing since for the default project settings the _process runs at the same pace.

https://phabricator.endlessm.com/T35563


addons/block_code/blocks/communication/area2d_on_entered.tres
```diff
@@ -10,8 +10,8 @@ description = ""
 category = "Communication | Methods"
 type = 1
 variant_type = 0
-display_template = "On [body: OBJECT] entered"
-code_template = "func _on_body_entered(body: Node2D):
+display_template = "When entering collision with [something: OBJECT]"
```
wjt, line 13: I don't like the "entering collision with" / "leaving collision with" pair. I don't think collisions are something you leave. It's awkward to phrase this because we don't have a [subject][]

[subject]: https://en.wikipedia.org/wiki/Subject_(grammar)

Brainstorming:

- When colliding with [something] / When no longer colliding with [something]
- When this node collides with [something] / When this node stops colliding with [something]
- When collision with [something] starts / When collision with [something] ends
- When [something] collides with this node / When [something] stops colliding with this node

Not sure I like these much better…

addons/block_code/blocks/lifecycle/process.tres
```diff
@@ -5,11 +5,12 @@
 [resource]
 script = ExtResource("1_pmina")
 name = &"process"
+target_node_class = ""
 description = "Attached blocks will be executed during the processing step of the main loop"
 category = "Lifecycle"
 type = 1
 variant_type = 0
-display_template = "On Process"
+display_template = "Every frame"
```
wjt, line 13: Are we going to consistently use *Sentence case* rather than *Title case* for blocks?

I checked what MakeCode Arcade does and actually it uses *lower case*!

addons/block_code/blocks/communication/area2d_on_entered.tres
dylanmccall: I like the second one of those; "When this node collides with / stops colliding with [something]". Alternatively, we could write these in the first person, like "When I collide with", which I think is the nicest to read here but stylistically it's tricky because we don't want to be writing _everything_ in first person.

addons/block_code/blocks/lifecycle/process.tres
manuq: Yes, I proposed lower case once to the other Will. Today we have a mix of Title case for short labels ("Viewport Center") and Sentence case for long labels. I just checked and Scratch has all in lower case too. So unless anyone has a strong argument I'll switch all to lower case.

addons/block_code/blocks/communication/area2d_on_entered.tres
manuq: This is a good point, should we switch everything to first person? Ie. "Remove me" instead of "Remove"? I'm open to do it.

manuq:
Ready for review again. I think that the switch to lower case makes the label much more readable inside blocks.

addons/block_code/blocks/communication/is_node_in_group.tres
```diff
@@ -5,11 +5,13 @@
 [resource]
 script = ExtResource("1_5krrs")
 name = &"is_node_in_group"
+target_node_class = ""
 description = "Is the node in the group"
 category = "Communication | Groups"
 type = 3
 variant_type = 1
-display_template = "Is {node: OBJECT} in group {group: STRING}"
+display_template = "is {node: OBJECT} in group {group: STRING}"
```
wjt, line 13: How about:

```suggestion
display_template = "{node: OBJECT} is in group {group: STRING}"
```

but this can be a follow-up.

addons/block_code/blocks/input/characterbody2d_move.tres
```diff
@@ -10,7 +10,7 @@ description = "Move the character up, down, left, and right with the keyboard us
 category = "Input"
 type = 2
 variant_type = 0
-display_template = "Move with keys {up: STRING} {down: STRING} {left: STRING} {right: STRING} with speed {speed: VECTOR2}"
+display_template = "move with keys {up: STRING} {down: STRING} {left: STRING} {right: STRING} with speed {speed: VECTOR2}"
```
wjt, line 13:
```suggestion
display_template = "move with keys {up: STRING} {down: STRING} {left: STRING} {right: STRING} at speed {speed: VECTOR2}"
```

addons/block_code/blocks/input/is_input_actioned.tres
```diff
@@ -17,7 +17,7 @@ description = "True if the specified input action has been pressed or released."
 category = "Input"
 type = 3
 variant_type = 1
-display_template = "Is action {action_name: STRING_NAME} {action: NIL}"
+display_template = "is action {action_name: STRING_NAME} {action: NIL}"
```
wjt, line 20:
```suggestion
display_template = "action {action_name: STRING_NAME} is {action: NIL}"
```

addons/block_code/blocks/loops/await_scene_ready.tres
```diff
@@ -5,11 +5,12 @@
 [resource]
 script = ExtResource("1_qy2t4")
 name = &"await_scene_ready"
+target_node_class = ""
 description = ""
 category = "Loops"
 type = 2
 variant_type = 0
-display_template = "Await scene ready"
+display_template = "await scene ready"
```
wjt, line 13:
```suggestion
display_template = "wait for the scene to be ready"
```

addons/block_code/blocks/math/randf_range.tres
```diff
@@ -10,7 +10,7 @@ description = "Generate a random floating point number between [i]from[/i] and [
 category = "Math"
 type = 3
 variant_type = 3
-display_template = "Random floating point number between {from: FLOAT} and {to: FLOAT}"
+display_template = "random floating point number between {from: FLOAT} and {to: FLOAT}"
```
wjt, line 13: A question for another day, but, do students know what "floating point number" means? Would "real number" be better understood?

addons/block_code/blocks/sounds/play_sound.tres
```diff
@@ -10,7 +10,7 @@ description = "Play the audio stream with volume and pitch"
 category = "Sounds"
 type = 2
 variant_type = 0
-display_template = "Play the sound {name: STRING} with Volume dB {db: FLOAT} and Pitch Scale {pitch: FLOAT}"
+display_template = "play the sound {name: STRING} with Volume dB {db: FLOAT} and Pitch Scale {pitch: FLOAT}"
```
wjt, line 13:
```suggestion
display_template = "play the sound {name: STRING} with volume {db: FLOAT} dB and pitch scale {pitch: FLOAT}"
```

`dB` is a unit, it goes after the value.

wjt (commented):
I think this would be fine but now all the block names are in the diff it's easier to review the others :)

dsd:
> I think this would be fine but now all the block names are in the diff it's easier to review the others :)

If/when we're looking at English language optimization can we check the animation player blocks text please (I think it had something like "play animation ahead" which can maybe be improved).

addons/block_code/blocks/input/characterbody2d_move.tres
manuq: I have to say that I'm not a fan of this block. I would prefer one block that can configure the player 1/2 keys, to work with the existing "move with player 1/2 buttons ...". But a discussion for another day.

addons/block_code/blocks/loops/await_scene_ready.tres
manuq: Definitely.

addons/block_code/blocks/math/randf_range.tres
manuq: I think that we should remove one of these random number blocks and have a single "random number between X and Y". What I'm not sure about is which one should survive, the one for floats or the one for ints. Let's do that as follow-up.

addons/block_code/blocks/sounds/play_sound.tres
manuq: Good point, and I forgot to lower case Volume and Pitch Scale.

manuq:
> > I think this would be fine but now all the block names are in the diff it's easier to review the others :)
> 
> If/when we're looking at English language optimization can we check the animation player blocks text please (I think it had something like "play animation ahead" which can maybe be improved).

Yes, is "play animation forward" better?

manuq:
@wjt @dylanmccall this is ready for review again.

PR #235, opened 2024-09-26
block: Show tooltips for blocks

wjt:
Prior to commit 7a6fe1329cb2d2b746a08fedb3fb6288cc0e6bba, instantiate_block() would set the new Block's tooltip_text property to the BlockDefinition's description.

Since that commit, we just set a definition property on the Block to the BlockDefinition.

Implement Control._get_tooltip(), returning BlockDefinition.definition, which restores tooltips to blocks.

https://phabricator.endlessm.com/T35645

### DIFF
--- a/addons/block_code/ui/blocks/block/block.gd
+++ b/addons/block_code/ui/blocks/block/block.gd
@@ -189,6 +189,10 @@ func _to_string():
 	return "<{block_class}:{block_name}#{rid}>".format({"block_name": definition.name if definition else "", "block_class": get_block_class(), "rid": get_instance_id()})
 
 
+func _get_tooltip(at_position: Vector2) -> String:
+	return definition.description if definition else ""
+
+
 func _make_custom_tooltip(for_text) -> Control:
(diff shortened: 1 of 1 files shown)


addons/block_code/ui/blocks/block/block.gd
```diff
@@ -189,6 +189,10 @@ func _to_string():
 	return "<{block_class}:{block_name}#{rid}>".format({"block_name": definition.name if definition else "", "block_class": get_block_class(), "rid": get_instance_id()})
 
 
+func _get_tooltip(at_position: Vector2) -> String:
+	return definition.description
```
dbnicholson, line 193: Unlikely, but I think to be safe you should check that `definition` is defined like in `_to_string`:

```suggestion
	return definition.description if definition else ""
```

dbnicholson (changes requested):
Nice! I was wondering how those were lost.

addons/block_code/ui/blocks/block/block.gd
wjt: Good point, I applied that change.

dbnicholson (approved):

PR #236, opened 2024-09-26
ParameterInput: Ignore mouse events when not visible

dylanmccall:
Instead of using only _input to process events, use _gui_input to watch for mouse button events that are understood to be for an input control, but use _input to watch for mouse movement events which may occur outside of the control.

To work within the rules of Godot's GUI event system, we need to do this in a control which is a child of each input control. This is achieved using draggable_overlay.tscn.

### DIFF
--- a/addons/block_code/ui/blocks/utilities/drag_drop_area/drag_drop_area.gd
+++ b/addons/block_code/ui/blocks/utilities/drag_drop_area/drag_drop_area.gd
@@ -1,16 +1,65 @@
 @tool
-extends MarginContainer
+## Drag drop area.
+##
+## A Control which watches for click and drag gestures beginning from itself.
+## It propagates events up to its parent, so it is possible to place this
+## control inside a control which processes input events such as [LineEdit].
+## If a drag occurs, it emits [signal drag_started].
+extends Control
 
-signal mouse_down
-signal mouse_up
+const Constants = preload("res://addons/block_code/ui/constants.gd")
+
+signal drag_started
+
+## True to require that the mouse move outside of the component before
+## [signal drag_started] is emitted.
+@export var drag_outside: bool = false
+
+var _drag_start_position: Vector2 = Vector2.INF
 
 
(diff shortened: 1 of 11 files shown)


addons/block_code/ui/blocks/utilities/drag_drop_area/drag_drop_area.gd
```diff
@@ -1,16 +1,60 @@
 @tool
-extends MarginContainer
+## Drag drop area.
+##
+## A Control which watches for click and drag gestures beginning from itself.
+## It propagates events up to its parent, so it is possible to place this
+## control inside a control which processes input events such as [LineEdit].
+## If a drag occurs, it emits [signal drag_started].
+extends Control
```
dbnicholson, line 8: Any reason to change this from a `MarginContainer`? Maybe the margin was pointless in this case since we just want the rectangle for collecting input events.

addons/block_code/ui/blocks/utilities/drag_drop_area/drag_drop_area.gd
```diff
@@ -1,16 +1,54 @@
...
+
+
+func _input(event: InputEvent) -> void:
+	if not event is InputEventMouseMotion:
+		return
+
+	if _drag_start == Vector2.INF:
+		return
+
+	var motion_event: InputEventMouseMotion = event as InputEventMouseMotion
```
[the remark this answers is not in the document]
dylanmccall: Thanks! Fixed :)

addons/block_code/ui/blocks/utilities/drag_drop_area/drag_drop_area.gd
dylanmccall: Yeah, it just didn't quite vibe with me. I think in an earlier version of this we were actually putting things inside of a DragDropArea? At this point we aren't.

dbnicholson (approved):
Looks good!

PR #237, opened 2024-09-26
Block: Set z_index on focus

dylanmccall:
Increase the z_index of the focused block's content, but decrease the z_index of its snap areas. This ensures that the focused block is completely in view, including its outline.


dbnicholson (approved):
Makes sense!

PR #238, opened 2024-09-26
DragDropArea: Add offset to drag_started signal

dylanmccall:
In DragManager, use the offset to ensure the block being dragged is
positioned under the cursor based on where the user started dragging.

-----

Based on #236.


starnight (approved):
LGTM!  Thanks!

PR #239, opened 2024-09-26
Pong: Use call method in node

dylanmccall:
Instead of setting a group for the "SimpleScoring" node, use the new call_method_node block.


starnight (approved):
Nice!

PR #240, opened 2024-10-01
blocks: Add groups option block extension

dbnicholson:
Several blocks work with node groups. Rather than having the user enter the group, offer the current groups as options. This is trickier than expected since Godot offers no API to get the current groups. Finding them involves 2 parts:

* Get scene local groups by calling get_groups() on each node in the scene.

* Get project global groups (starting in 4.3) by getting the `global_group` values in the project settings.

Monitoring for changes to the global groups is possible by listening for project settings changes. Monitoring for local group changes is not possible since the only events that occur are internal to Godot and not available in GDScript.

https://phabricator.endlessm.com/T35647

This is a naive implementation. The groups are specific to the scene and project. Rather than have each `Block` maintain its own list, it should be gathered in the `BlockCode` node or even the `BlockCodePlugin`. I spent far too long trying to make that work well, but I decided to just put up this simple implementation now.

As noted above, the biggest problem is that the scene local groups are not monitored. So, if you add a node to a group, the Blocks won't be updated with the new group. Also, since the `add_to_group` blocks only offer existing groups, it takes away the means of creating scene local groups dynamically. That may not be that bad, though, since that would only happen at runtime and you wouldn't be able to query for the presence of the group.

### DIFF
--- a/addons/block_code/blocks/communication/groups.gd
+++ b/addons/block_code/blocks/communication/groups.gd
@@ -1,0 +1,71 @@
+@tool
+## Common BlockExtension for scene group options.
+extends BlockExtension
+
+const OptionData = preload("res://addons/block_code/code_generation/option_data.gd")
+const Util = preload("res://addons/block_code/ui/util.gd")
+
+
+# Global groups are just project settings in the global_group group.
+# ProjectSettings doesn't offer an API to get the project groups or to get all
+# settings. Fortunately, settings are simply implemented as properties on the
+# ProjectSettings object.
+static func _add_global_groups(groups: Dictionary):
+	for prop in ProjectSettings.get_property_list():
+		var name: String = prop["name"]
+		var parts := name.split("/", false, 1)
+		if parts[0] == "global_group":
+			groups[parts[1]] = null
+
+
+# Add all the groups in a node and its children, ignoring internal _ prefixed
+# groups.
+static func _add_node_groups(groups: Dictionary, node: Node):
+	for group in node.get_groups():
+		if not group.begins_with("_"):
+			groups[String(group)] = null
+
+	for child in node.get_children():
+		_add_node_groups(groups, child)
+
+
+func _get_edited_scene_groups() -> Array[String]:
+	var groups: Dictionary
+	_add_global_groups(groups)
+
+	var root := context_node.get_tree().edited_scene_root
+	_add_node_groups(groups, root)
+
+	var sorted_groups: Array[String]
+	sorted_groups.assign(groups.keys())
+	sorted_groups.sort()
+	return sorted_groups
+
+
+func _init():
+	# FIXME: Only global group changes are monitored. Scene local groups should
+	# also be monitored, but godot does not have any reasonable API to do that.
+	ProjectSettings.settings_changed.connect(_on_project_settings_changed)
+
+
+func _context_node_changed():
+	# If the context node changed, the scene local groups need to be updated.
+	changed.emit()
+
+
+func get_defaults() -> Dictionary:
+	if not context_node:
+		return {}
+
+	# The default groups are only needed in the editor.
+	if not Util.node_is_part_of_edited_scene(context_node):
+		return {}
+
+	var groups: Array[String] = _get_edited_scene_groups()
+	return {"group": OptionData.new(groups)}
+
+
+func _on_project_settings_changed():
+	# FIXME: The global groups should be cached and compared so that the
+	# defaults are only changed when the global groups actually change.
+	changed.emit()
--- a/addons/block_code/blocks/communication/is_in_group.tres
+++ b/addons/block_code/blocks/communication/is_in_group.tres
@@ -1,6 +1,13 @@
-[gd_resource type="Resource" load_steps=2 format=3 uid="uid://q4cnstftvsiu"]
+[gd_resource type="Resource" load_steps=5 format=3 uid="uid://q4cnstftvsiu"]
 
+[ext_resource type="Script" path="res://addons/block_code/code_generation/option_data.gd" id="1_cla3i"]
 [ext_resource type="Script" path="res://addons/block_code/code_generation/block_definition.gd" id="1_tjyq5"]
+[ext_resource type="Script" path="res://addons/block_code/blocks/communication/groups.gd" id="2_o165d"]
+
+[sub_resource type="Resource" id="Resource_d0v0d"]
+script = ExtResource("1_cla3i")
+selected = 0
+items = []
 
 [resource]
 script = ExtResource("1_tjyq5")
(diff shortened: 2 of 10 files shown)


starnight (approved):
LGTM!

However, `global_group` is supported since Godot 4.3, should we set the supported Godot to 4.3?

addons/block_code/blocks/communication/groups.gd
```diff
@@ -0,0 +1,70 @@
...
+		if name.begins_with("global_group/"):
+			var parts = name.split("/", false, 1)
+			groups[parts[1]] = null
+
+
+# Add all the groups in a node and its children, ignoring internal _ prefixed
+# groups.
+static func _add_node_groups(groups: Dictionary, node: Node, scene_root: Node):
+	# Ignore this node and its children if it isn't part of the scene.
+	if node != scene_root and node.owner != scene_root:
```
dylanmccall, line 25: One issue with this use of `node.owner` is we're going to skip nodes in an instantiated scene. For instance, with our pong_game.tscn example, the BlockCode for Ball refers to a group named "walls", but that group isn't in the generated options list because we created that group in the separate Space scene. Was this intentional? Otherwise, it looks to me what we want here is more like...

```suggestion
	if node != scene_root and not scene_root.is_ancestor_of(node):
```

dylanmccall (changes requested):

addons/block_code/blocks/communication/groups.gd
dbnicholson: That was intentional. Since that's a different scene, it didn't seem appropriate to include it's groups since that's not how Godot itself handles them. Hmm, actually I guess that's not true testing it out. After adding a group in a child scene, the Groups Editor shows it as available in the parent scene but with a :lock: and a tooltip saying "This group belongs to another scene and can't be edited".

If we're just going to descend all the child scenes then we can just drop this conditional completely. We already know it's an ancestor since the function is walking children recursively. Lemme try that out...

dbnicholson:
I now noticed that if I change the edited scene, the groups shown in the picker block won't change until I start dragging it out of the picker. I think I have to watch for the context node changing and emit BlockExtension.changed.

dylanmccall:
> LGTM!
> 
> However, `global_group` is supported since Godot 4.3, should we set the supported Godot to 4.3?

I think we're okay there. I tested with Godot 4.2.2 to be sure, and it still works as expected except for an unrelated issue with simple_spawner.gd. Groups defined in the scene file still appear in the block, although global groups _are_ pretty helpful in this case.

dbnicholson:
> > LGTM!
> > However, `global_group` is supported since Godot 4.3, should we set the supported Godot to 4.3?
> 
> I think we're okay there. I tested with Godot 4.2.2 to be sure, and it still works as expected except for an unrelated issue with simple_spawner.gd. Groups defined in the scene file still appear in the block, although global groups _are_ pretty helpful in this case.

Indeed. Since the implementation is just to scrape them out of the ProjectSettings properties, all that will happen on 4.2 is that there are never global groups.

dbnicholson:
> I now noticed that if I change the edited scene, the groups shown in the picker block won't change until I start dragging it out of the picker. I think I have to watch for the context node changing and emit BlockExtension.changed.

Actually, this is an existing problem in the picker. When the context has changed and blocks are being reused, the context node change needs to be passed down to the block extension. I think I have something working.

dbnicholson:
I changed a few things:

* Groups from all children of the edited scene root are included. Previously only nodes in the scene were considered, but that doesn't match the Groups Editor, which shows groups from child scenes.
* When the extension's context node changes, its changed signal is emitted so that consumers will refresh the defaults based on the node's scene.
* I added a commit that updates the extension's context node from the picker since it reuses blocks and the options wouldn't be updated.

addons/block_code/ui/blocks/block/block.gd
```diff
@@ -88,6 +88,12 @@ func get_parameter_values() -> Dictionary:
 	return template_editor.get_parameter_values()
 
 
+## Use the current BlockEditorContext components
+func refresh_context():
+	if _context and _block_extension:
+		_block_extension.context_node = _context.parent_node
```
dylanmccall, line 94: Can this be a signal handler for `_context.changed`, instead, or does that do something werid? (Oh noo so many changed signals!). Alternatively I don't mind if we move this stuff over to BlockExtension itself. For BlockExtension I didn't want to add anything more than that one context_node property until it was needed, but as long as individual extensions aren't poking at BlockEditorContext it's fine.

dbnicholson: That was my initial instinct, but that's not what we want. If the BlockExtension watched for context changes, then the options would change every time you changed the block code node. We only want to do this in the picker where the blocks are reused for all contexts. I presume that's why you passed in the context node to BlockExtension instead of having it just grab the current context.

dylanmccall: Oh, yeah, I see where you're coming from there. Thanks for the explanation!

dylanmccall:
This looks good to me! I'll merge it now :)

addons/block_code/blocks/communication/is_in_group.tres
```diff
@@ -10,8 +17,11 @@ description = "Is this node in the group"
...
 type = 3
 variant_type = 1
-display_template = "is in group {group: STRING}"
-code_template = "is_in_group({group})"
-defaults = {}
+display_template = "is in group {group: NIL}"
+code_template = "is_in_group(\\\"{group}\\\")"
+defaults = {
+"group": SubResource("Resource_d0v0d")
+}
```
dylanmccall, line 24: Ack, I completely missed this change! We shouldn't have needed to change group to a NIL :( I'll make a PR fixing this in a moment.

dylanmccall: To explain what's going on, _previously_ we needed to do that for options lists because the value was just reproduced verbatim, but now it can be treated as a string and auto-quoted just like any other string parameter. The old behaviour was used in places, and that can be achieved by using something like `{{param}}` in the code template to substitute just the raw value.

dbnicholson: What was the change that made that different? I've been basing my work on 41afa00261e17606c0754b8697940f65c79c0194 where several string types changed to `NIL`.

dylanmccall: Oh, I see what happened there! Some crossed wires :) I think I switched them to NIL to keep the previous behaviour of not being able to snap weird things to options lists, but then #219 happened around the same time so it was kind of moot, and the parameter type is indeed pretty irrelevant to anything. The only thing that's important is we don't need to do anything special to quote the strings here.

The thing that caused the behaviour change in that commit is where we remove the special case for `TYPE_OBJECT` in `raw_input_to_code_string` (seeing as option values are no longer `OptionData` objects).

Issue #184, opened 2024-08-02
Scrolling or panning in any bottom panel panel is incorrectly handled by the Block Code panel


dylanmccall:
### What Happened

In https://github.com/endlessm/godot-block-coding/pull/170#issuecomment-2259058174, @manuq pointed out…

> Regarding focus, something is wrong in both main and this branch: when I scroll in the Output dock the Block Code canvas is zoomed in and out. As you can see in this video:
[Grabación de pantalla desde 2024-07-29 14-44-24.webm](https://github.com/user-attachments/assets/3c4fe60d-87ac-4ce7-b1c9-7178c1801c2b)

### How to Reproduce

1. In the Block Code panel, open a Block Code script.
2. Switch to the Output panel.
3. In the Output panel, scroll down using the mouse wheel, or click and drag with the middle mouse button.
4. Switch back to the Block Code panel.
5. Note that the view has been panned and zoomed out.

### Block Coding Plugin Version

v0.6.2

### Godot Engine Version

v4.2.2.stable

### Operating System

Fedora Silverblue 40

PR #241, opened 2024-10-01
BlockCanvas: Use _gui_input instead of _input

dylanmccall:
Use the _gui_input virtual function to pan and zoom the block canvas, instead of _input. This way, events are filtered according to the block canvas's visibility. As a side-effect, the block canvas processes input events after any containing widgets. For example, middle clicking inside a parameter input component will no longer begin panning. However, this is a better situation than the block canvas handling input events regardless of its visibility.

Closes: https://github.com/endlessm/godot-block-coding/issues/184


dbnicholson (approved):
I really wanted to do this before but I was too scared of the implications. I trust you do know the implications, though!

PR #242, opened 2024-10-01
Fix error in simple_spawner.gd with Godot 4.2.2

dylanmccall:
With Godot 4.3, it is possible to use is_instance_valid as a Callable. To support Godot 4.2.2, we must use an anonymous function instead.

### DIFF
--- a/addons/block_code/simple_spawner/simple_spawner.gd
+++ b/addons/block_code/simple_spawner/simple_spawner.gd
@@ -79,7 +79,7 @@ func spawn_once():
 	if scenes.size() == 0:
 		return
 
-	_spawned_scenes = _spawned_scenes.filter(is_instance_valid)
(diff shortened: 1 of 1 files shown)


addons/block_code/simple_spawner/simple_spawner.gd
```diff
@@ -79,7 +79,7 @@ func spawn_once():
 	if scenes.size() == 0:
 		return
 
-	_spawned_scenes = _spawned_scenes.filter(is_instance_valid)
+	_spawned_scenes = _spawned_scenes.filter(func(instance): return is_instance_valid(instance))
```
starnight, line 82: I try to understand the `func()` and `instance`.  Then, I found [Godot 4 lambda functions](https://www.youtube.com/watch?v=N4yX8lbBRWY).

starnight (approved):

PR #243, opened 2024-10-03
blocks: Use string type for group names

dylanmccall:
This spares us some grief in the code templates for different group-related blocks.

https://phabricator.endlessm.com/T35647


dbnicholson (approved):
Oh, I thought the type always had to be `NIL` when using options.

PR #244, opened 2024-10-04
pong_game: Use SimpleSpawner and global groups

dylanmccall:
Instead of resetting the ball, use a SimpleSpawner. This explores the use of BlockCode nodes across multiple scenes and replaces a use of call_method_group with call_method_node.

-----

I'm not sure if we really want this, so I'll mark it as a draft even though I'm finished poking at it.


manuq:
Yes, I think this makes the demo game better.

PR #246, opened 2024-10-08
Update screenshot

wjt:
The previous version of the screenshot showed a much older plugin version. Update it to show the same program – the code for a paddle in Pong – but in the new split view with the 2D canvas above it.

This will be used on the Godot asset library, though updating it is (currently) a manual process.


wjt:
Before:



Now:

cassidyjames (approved):
:shipit:

PR #250, opened 2024-10-09
simple_scoring: Show a single score label

dylanmccall:
Instead of SimpleScoring providing a left and right score, change it to keep track of a single score, and show a single label. In the pong_game scene, we add two nodes: one for the left player, one for the right player.

### DIFF
--- a/addons/block_code/simple_nodes/simple_scoring/simple_scoring.gd
+++ b/addons/block_code/simple_nodes/simple_scoring/simple_scoring.gd
@@ -1,48 +1,32 @@
 @tool
 class_name SimpleScoring
-extends CanvasLayer
+extends Node2D
 
 const BlockDefinition = preload("res://addons/block_code/code_generation/block_definition.gd")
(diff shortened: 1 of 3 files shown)


addons/block_code/simple_nodes/simple_scoring/simple_scoring.gd
```diff
@@ -1,48 +1,32 @@
...
 
-const _POSITIONS_FOR_PLAYER = {
-	"1": "left",
-	"2": "right",
-}
+func _create_score_label():
+	if _score_label:
+		return
 
+	_score_label = Label.new()
```
dbnicholson, line 19: Why not just make this extend Label instead of having a child label?

dylanmccall: I played with that :) My reason here is I felt it's a _feature_ that SimpleScoring provides a simple interface in the editor: no label properties, no extra Control-related handles in the canvas. But with that in mind, I created #249 to add a block for setting a label's text.

dbnicholson (approved):

Issue #248, opened 2024-10-09
how ican use velocity.x it self 


wesamdev:
### What Happened

i cant make player movement with jump correctly because i cant set velocity.x it self without set y to 0****

### How to Reproduce

go to velocity tab and add set velocity

### Block Coding Plugin Version

v0.7.1 Latest

### Godot Engine Version

v4.4.0-dev redot

### Operating System

windows 10 pro x64

dylanmccall:
Fixed in 877f5625f29c4e6210ade9afd4194a1ba3abd36e.

dylanmccall:
Thank you for your suggestion! :)

wesamdev:
thanks for the fix

wesamdev:
> Thank you for your suggestion! :)

please add it also for velocity

PR #253, opened 2024-10-10
Vector2 x and y blocks

DoomTas3r:
Adds the `x of (Vector2)` and `y of (Vector2)` blocks. The blocks are not ordered in the block picker

Helping with the issue https://github.com/endlessm/godot-block-coding/issues/248#issue-2575803628

image

image

- Using the `(string1)+(string2)` block adds numbers instead of concatenating. This is probably a feature of the GDScript `+` operator. A separate concatenation block or a number to string block may need to be added

### DIFF
--- a/addons/block_code/blocks/math/vector2_x.tres
+++ b/addons/block_code/blocks/math/vector2_x.tres
@@ -11,7 +11,7 @@ category = "Math"
 type = 3
 variant_type = 3
 display_template = "x of {vector2: VECTOR2}"
-code_template = "({vector2}).x"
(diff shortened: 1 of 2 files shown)


dylanmccall:
Thank you so much for your pull request @DoomTas3r! We're actually in the middle of a little game jam ahead of GodotCon and literally needed to add these exact blocks :laughing:

It looks good to us. All good if we go ahead and merge it?

dylanmccall:
I squashed and merged this into main just now, but I'll leave this PR open for the moment in case there was something else you intended to change :)

PR #255, opened 2024-10-10
Expose _process()'s delta parameter as block

wjt:
This might be useful in some programs.

Draft because I don't like the text on the block.


wjt:
image

wjt:
IRL @dsd reckons that, regardless of text, the added complexity of giving this block a parameter is not worth it: this is a block that every learner will need to use.

He suggested we could have 2 blocks:

- every frame
- every frame, i'm an expert, with delta (text tbc)

PR #258, opened 2024-10-10
Await until animation is finished

manuq:
Add an option for playing the animation and waiting until it's done. The previous behavior is now another option "play in the background".

For editing these block definition options in the Inspector, the class name is back to OptionData.


dylanmccall (approved):
I fixed a small issue where `wait_mode` was wrapped in the wrong kind of braces, and removed a change adding `class_name OptionData` to option_data.gd. I do think we should do that, but I don't recall why we don't at the moment and I think we should change it in a more purposeful place.

Issue #261, opened 2024-10-11
add physics process


wesamdev:
### What Happened

please add physics process loop

### How to Reproduce

nothing

### Block Coding Plugin Version

v0.7.1 Latest

### Godot Engine Version

v4.4.0-dev redot

### Operating System

windows 10 pro x64

manuq:
Thanks. We had this before and it was intentionally removed, because it is a bit confusing for learners to have two different main loops. And because for the default Godot project, the physics loop and the screen refresh loop run at the same pace.

manuq:
Here is a proposal to make this happen: #282

manuq:
Thanks for your contribution, @DoomTas3r .

PR #262, opened 2024-10-11
Physics process block

DoomTas3r:
Adds the `every physics step` block. The [documentation](https://docs.godotengine.org/en/stable/tutorials/physics/physics_introduction.html#physics-process-callback) says that the `_physics_process()` function is native to every node and is called before each physics step. The debugger displays a warning because `delta` is unused

Helping with the issue https://github.com/endlessm/godot-block-coding/issues/261#issue-2580370059 (This might not fix the issue if the `delta` parameter is needed)
image

image

- Should the description be more informative?


manuq:
This will be welcome once #261 is implemented.

manuq:
This was merged in #342 .

PR #263, opened 2024-10-11
block_canvas: Select the new block code node in add block code

dylanmccall:
This fixes an issue where the blocks picker incorrectly generates its list of blocks before the block code node's block script is ready.


dbnicholson (approved):
Thanks! For reference, this was broken by 58983d0cd580c9bf2858ed379e77520c0d7fcde4.

Issue #131, opened 2024-07-12
Block Picker takes up too much space


JohnsonMyton:
### What Happened

Hello! Love the idea of the project, wanted to add one of my first, instant thoughts. The amount of space for the actual blocks is pretty small, and on my students computers it's even smaller. One of the issues I think is that the gifts take up too much space - 
![image](https://github.com/user-attachments/assets/629aff86-5cf4-4a43-bbe9-571920f53524)

### Spec

The Block Picker should be a single area. Currently it has 2 columns: one for the categories (left) and one for the blocks themselves (right).

Convert the category selector buttons (currently first column) into title bar buttons. When clicking a title bar button, the page scrolls to show the blocks of that category. There is no need for subcategory buttons (eg. Logic | Conditionals), just category (eg. Logic).

The title bar buttons should be icon only, with a tooltip for the category name. Each category must have its own distinguishable icon. And colored with the category color.

#### References
- Use this mockup (left image) for reference: https://github.com/endlessm/godot-block-coding/issues/131#issuecomment-2485868625
- Use icons from https://developer.gnome.org/hig/guidelines/ui-icons.html or provide custom ones. They must be imported to the project as SVG files.

dsd:
@JohnsonMyton what's the specs of the students machines? And the screen resolution?

Excited to hear you are considering this for use in education context. Could you reach out to me at drake@endlessos.org - I would love to know more of the context. We're also developing this for learners but haven't manged to expose many students to this yet because of summer break.

cassidyjames:
An easy change here might be moving the category to the end of the line… so you'd have:

```
Lifecycle
Position | Transform
Rotation | Transform
Scale | Transform
Modulate | Graphics
Visibility | Graphics
Viewport | Graphics
Sounds
Input
```

Then users who collapse the width of that column can still see the most useful bit.

If we want to get more interesting, we could consider relying on an icon + color to group them into their respective groups instead, dropping the category from the label (and/or moving it to a tooltip). This would allow us to better use the space while still keeping a pretty strong association between blocks of the same category.

The [GNOME UI icons](https://developer.gnome.org/hig/guidelines/ui-icons.html) are public domain, so we could lean on those for each category.

manuq:
@cassidyjames what do you think about the implementation done by @DoomTas3r here? #270

I have tried it and although it helps, I still find odd to have 2 columns in the tab:
[...]

PR #270, opened 2024-10-13
Custom category icons, tooltips, and removed labels

DoomTas3r:
Removes category labels and subcategories, and adds tooltips and custom icons. Loading the icons was not optimized

Based on the idea https://github.com/endlessm/godot-block-coding/issues/131#issuecomment-2253548713

image


DoomTas3r:
I think I understand how squash works now. I will squash the commits in other pull requests

cassidyjames:
On HiDPI (2× scaling) the icons look tiny; is there a way to make them scale with the UI? 
image

DoomTas3r:
@cassidyjames I changed the import files so that the icons scale with the editor. How do they look now?

manuq:
@DoomTas3r this is looking very nice! By the way, I don't have an HiDPI display.
Captura desde 2025-02-03 12-20-20

cassidyjames (approved):
UI-wise this looks and feels great to me. Nice work

image


I've confirmed the icons now scale correctly on HiDPI. I was going to point out that it looks like the category labels don't scale, but they don't in main, either…

image

I'll file a separate issue for that. :)

manuq:
@DoomTas3r would you like to cleanup the PR? The second commit undo things done by the first one. For example I see icons added and then removed. So you should squash them into one and provide a descriptive commit message. Then we can start reviewing!

addons/block_code/ui/constants.gd
```diff
@@ -24,130 +24,156 @@ const BUILTIN_CATEGORIES_PROPS: Dictionary = {
 	{
 		"color": Color("ec3b59"),
 		"order": 10,
+		"icon": "lifecycle",
```
manuq, line 27: Below in the PR you do this for the button tooltip text:
```
_button.tooltip_text = tr(category.name.get_slice(" |", 0))
```

So I think it's better to do the same thing instead of introducing a new "icon" key in this constant. If you want to keep the SVG files lowercase, you can use [to_lower()](https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-to-lower)

manuq (changes requested):

DoomTas3r:
@manuq Alright, it's all cleaned up!

manuq (approved):

PR #272, opened 2024-10-14
SimpleSpawner: Rename frequency to period

wjt:
This property is the time, in seconds, to wait between spawning scenes. This is a period/interval/wavelength, not a frequency, which would be measured in Hz, or 1/seconds.

Adjust the name of the property and its documentation.

The instances of “frequency” that I have not changed are in the example scene's on-screen documentation. In this case it was actually used correctly: the keyboard action to increase the frequency reduces the property which is now called period, and the decrease action respectively increases the period.

I also added an annotation I learned yesterday to put units in the inspector slider for spawn period and spawn limit.

Screenshot from 2024-10-14 11-21-18

### DIFF
--- a/addons/block_code/examples/spawner/spawner.tscn
+++ b/addons/block_code/examples/spawner/spawner.tscn
@@ -37,7 +37,7 @@ arguments = {
 
 [sub_resource type="Resource" id="Resource_uv0fo"]
 script = ExtResource("7_cykhe")
-name = &"simplespawner_get_spawn_frequency"
+name = &"simplespawner_get_spawn_period"
 arguments = {}
 
 [sub_resource type="Resource" id="Resource_l45hk"]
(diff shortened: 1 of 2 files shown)


dbnicholson (commented):
I agree in principle with this change completely even though I doubt learners will have the nuance to differentiate the terms or even know what they are. That said, I think we do want to use the correct terms.

An issue here is backwards compatibility. Changing the exported variable and block definition names will break any scene using SimpleSpawner (as implied by the changes in the example spawner scene). While we have not committed to any backwards compatibility, I feel like we shouldn't put this in a 0.7 patch release.

I think it would be possible to maintain compatibility in a couple ways.

1. Keep an unexported `spawn_frequency` that wraps `spawn_period` with getters and setters.
2. Duplicate the block definitions with the old names. However, that would make both blocks show up in the picker. You'd probably need to add some type of `hidden` block definition attribute. Or you could add an `aliases` block definition attribute that was used for lookup when loading.

I'm not sure we need to bother, but I wanted to discuss before merging.

manuq:
Yes I think it's worth discussing a migration path. The alternatives 1. and 2. don't sound to me a proper migration, but more an effort to keep backwards compat with some overhead. Still we are considering this plugin as experimental and in early stages (as said in the README). So I'd say let's merge this for now.

wjt:
We came to the conclusion in discussion today that we should instead remove this demo because the new one from godotcon is a better demo of how to make a runner.

manuq:
@wjt maybe you wanted to close https://github.com/endlessm/godot-block-coding-demos/pull/13 instead?

wjt:
Oops, yes I did. Well noticed!

wjt:
I think I'll take a quick look at compatibility..

dbnicholson:
Let's just do this and ignore the compatibility for now.

PR #273, opened 2024-10-14
Adds hexagons

DoomTas3r:
Blocks can now make a wish to become 1 of 4 variants of hexagons. Blocks can become pure hexagons. Outlines are optional

image

image


manuq:
Hi @DoomTas3r. Although nice, adding variants to the blocks shape is not fixing or improving anything per se, so unlikely to be merged. But here is a proposal to improve the entry blocks and your help would be very appreciated! https://github.com/endlessm/godot-block-coding/issues/278

Can you retarget this PR to make that improvement?

manuq:
@DoomTas3r this is another issue that you can help with related to block shapes: https://github.com/endlessm/godot-block-coding/issues/279

DoomTas3r:
@manuq I think this should be good. You can tell me if any errors related to this are found. The other types can also be styled but I was hesitant because confusion can be caused from overloading

manuq:
@DoomTas3r whoa!! What you managed to do regarding shapes is amazing, thanks! I'll review the code in a bit.

addons/block_code/ui/blocks/parameter_block/parameter_block.gd
```diff
@@ -18,16 +19,22 @@ var _panel_focus: StyleBox
 func _ready():
 	super()
 
-	_panel_normal = _panel.get_theme_stylebox("panel").duplicate()
-	_panel_normal.bg_color = color
-	_panel_normal.border_color = color.darkened(0.2)
+	if not definition == null and definition.variant_type == Variant.Type.TYPE_BOOL:
+		_background.visible = true
+		_background.variant = 3
```
manuq, line 24: Please use [enumerators](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_basics.html#enums) rather than integers. For example:

```
enum BackgroundVariant { ROUND, POINTY, ... }
```

addons/block_code/ui/blocks/utilities/background/background.gd
```diff
@@ -21,18 +27,36 @@ var parent_block: Block
 @export var shift_bottom: float = 0.0:
 	set = _set_shift_bottom
 
+## Style of the top knob
+@export var top_variant: int = 0:
+	set = _set_top_variant
+
+## |0|, \1/, /2/, <3>, >4>, v5v, v6^, \7y, /8y
```
manuq, line 34: We really don't need the 9 shapes that the first commit adds, and they make the code very hard to read. We only need:
- A pointy background for value blocks of variant type BOOLEAN.
- A round ellipse for entry blocks.

addons/block_code/ui/blocks/utilities/background/background.gd
```diff
@@ -43,67 +67,196 @@ func _set_shift_bottom(new_shift_bottom):
...
+		fill_polygon.append_array(top_knob)
 
-	var stroke_polygon: PackedVector2Array
-	var edge_polygon: PackedVector2Array
-	var outline_middle = Constants.OUTLINE_WIDTH / 2
+	# Right side
+	if variant > 0:
+		# Top
+		if variant == 3 or variant == 4:
+			fill_polygon.append(Vector2(size.x - 5.0, 0.0))
```
manuq, line 129: These `5.0` are hardcoded everywhere. Please add a new "pointy width" constant:
```suggestion
			fill_polygon.append(Vector2(size.x - Constants.POINTY_WIDTH, 0.0))
```

manuq (changes requested):
@DoomTas3r I've been delaying this review so much, sorry! Here are a few changes requested. It works like a charm.

DoomTas3r:
@manuq Okay, the code was cleaned up. I thought maybe the Vector2 casting would be confusing, but I checked this block configuration
image
and it gave the error `Parser Error: Value of type "Vector2" cannot be assigned to a variable of type "bool".` which I thought was possible according to https://github.com/endlessm/godot-block-coding/pull/254 and https://docs.godotengine.org/en/stable/classes/class_vector2.html but it seems that `Vector2` can only be used as a `bool` for operations because this configuration works
image

manuq:
@DoomTas3r thanks! The PR diff is much clear now. Now you can see that the last commit reverts things added in the previous ones. The next level in contribution is to do an interactive rebase of your work using `git rebase --interactive` on your end. Then you could (for example) come up with two clean commits:
- Restyle entry shape
- Make boolean blocks shape pointy

For the sake of merging this soon, I'll do the interactive rebase myself this time.

PR #276, opened 2024-10-18
blocks: Remove extra newlines from entry blocks

dbnicholson:
Entry blocks already add a trailing newline, so there's no need to include one in the code template. See the ready, process and define_method blocks, for example. Having one doesn't cause any harm, but it does make the generated code uglier.


wjt:
Funny, the tests are asserting that the ugly newlines are present...

dbnicholson:
> Funny, the tests are asserting that the ugly newlines are present...

Whoops! Forgot to check the tests...

wjt (approved):

PR #277, opened 2024-10-18
simple_nodes: Add SimpleEnding node

dbnicholson:
This provides a label that displays either a win or lose message along with blocks to set or clear the label during the game. Label is inherited directly so that all the label properties are available in the inspector.

https://phabricator.endlessm.com/T35661

### DIFF
--- a/addons/block_code/simple_nodes/simple_ending/simple_ending.gd
+++ b/addons/block_code/simple_nodes/simple_ending/simple_ending.gd
@@ -1,0 +1,89 @@
+@tool
+class_name SimpleEnding
+extends Label
+
+const BlockDefinition = preload("res://addons/block_code/code_generation/block_definition.gd")
+const BlocksCatalog = preload("res://addons/block_code/code_generation/blocks_catalog.gd")
+const OptionData = preload("res://addons/block_code/code_generation/option_data.gd")
+const Types = preload("res://addons/block_code/types/types.gd")
+
+## The message that will be shown when the player wins the game.
+@export var win_message: String = "YOU WIN!"
+
+## The message that will be shown when the player loses the game.
+@export var lose_message: String = "GAME OVER"
+
+
+func game_over(result: String):
+	match result:
+		"WIN":
+			text = win_message
+		"LOSE":
+			text = lose_message
+		_:
+			text = ""
+			push_warning('Unrecognized game result "%s"' % result)
+
+
+func reset():
+	text = ""
+
+
+func _ready():
+	simple_setup()
+
+
+func simple_setup():
+	if not label_settings:
+		label_settings = LabelSettings.new()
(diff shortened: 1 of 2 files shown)


addons/block_code/simple_nodes/simple_ending/simple_ending.gd
```diff
@@ -0,0 +1,80 @@
...
+	simple_setup()
+
+
+func simple_setup():
+	if Engine.is_editor_hint():
+		# Show the win message in the editor so adjusting the label properties
+		# is visible.
+		text = win_message
+	else:
+		text = ""
```
manuq, line 42: Nit: this makes it not work when starting, even adding the "wait for scene to be ready" in the middle:
Captura desde 2024-10-18 15-40-21


```suggestion
``

manuq (changes requested):
Nice, I like how it inherits Label so things like font color can be tweaked (if you can find them).

Only a nit change requested.

addons/block_code/simple_nodes/simple_ending/simple_ending.gd
dbnicholson: It is broken if you try to run it from `_ready` since `simple_setup` is run after the node is ready. The problem this is trying to address is that if you set the label text in the editor, it will be serialized to the scene and restored when loading. I tried to figure out how to clear the label text when saving so that it didn't get serialized, but I couldn't figure it out.

I think I'm just going to change this so that SimpleEnding descends from Control and the Label is created at runtime.

manuq: > Updated now so the Label isn't saved to the scene. Since the Label is created in `simple_setup` and that's not run until the scene becomes idle, I had to add a signal and await it to make the blocks work from the `when starting` block.

Hmm I really liked it when it was inheriting Label, because it allowed the user to change the font, text color, etc. I was even going to propose this inheritance for SimpleScoring. I wasn't expecting a reimplementation like this for that little nit observation :) . Can you go back to it? Is actually unlikely that someone would like to set "you win" or "you lose" when starting.

dbnicholson: A LabelSettings instance is still exposed in this case, but I can try to go back.

Possibly what I can do is toggle the node visibility from `enter/exit_tree` when not in editor mode.

dbnicholson: I changed it back to a Label but moved the hack to clear the text to `_enter_tree` when not in the editor. That seems to work well enough to use the game over blocks from "when starting".

manuq: Excellent!

PR #283, opened 2024-10-22
SimpleSpawner: Rename frequency to period (with backwards compatibility)

wjt:
This is an alternative to:

- #272

with backwards compatibility with existing block programs. It roughly implements @dbnicholson's suggestion in https://github.com/endlessm/godot-block-coding/pull/272#pullrequestreview-2369649749:

> An issue here is backwards compatibility. Changing the exported variable and block definition names will break any scene using SimpleSpawner (as implied by the changes in the example spawner scene). While we have not committed to any backwards compatibility, I feel like we shouldn't put this in a 0.7 patch release.
> 
> I think it would be possible to maintain compatibility in a couple ways.
> 
> 1. Keep an unexported `spawn_frequency` that wraps `spawn_period` with getters and setters.

Actually it has to be exported; otherwise values in existing `.tscn` files are not set. I learned about the [`@export_storage`](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_exports.html#export-storage) annotation, added in Godot 4.3, which does what we want here. However that does mean bumping the minimum supported Godot version. And also it means that both properties will be saved in scenes – not only existing scenes which instantiated SimpleSpawner and set the old property, but also those that set the new name.

> 2. Duplicate the block definitions with the old names. However, that would make both blocks show up in the picker. You'd probably need to add some type of `hidden` block definition attribute. Or you could add an `aliases` block definition attribute that was used for lookup when loading.

I took the `hidden` path. But having written it, I think the result is kind of ugly. Maybe `aliases` would be better.

It was interesting to learn some things about properties, but having written all this I kind of think we should just go with #272: make the destructive change and call the next release 0.8.0. I'm interested in other opinions though.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -27,7 +27,7 @@ jobs:
     name: Tests
     strategy:
       matrix:
-        godot-version: [4.2.2, 4.3.0]
+        godot-version: [4.3.0]
     runs-on: ubuntu-latest
     steps:
     - name: Checkout
--- a/addons/block_code/blocks/communication/add_node_to_group.tres
+++ b/addons/block_code/blocks/communication/add_node_to_group.tres
@@ -25,3 +25,4 @@ defaults = {
 signal_name = ""
 scope = ""
 extension_script = ExtResource("1_p83c7")
+hidden = false
--- a/addons/block_code/blocks/communication/add_to_group.tres
+++ b/addons/block_code/blocks/communication/add_to_group.tres
@@ -25,3 +25,4 @@ defaults = {
 signal_name = ""
 scope = ""
 extension_script = ExtResource("2_42ixf")
+hidden = false
--- a/addons/block_code/blocks/communication/call_method_group.tres
+++ b/addons/block_code/blocks/communication/call_method_group.tres
@@ -25,3 +25,4 @@ defaults = {
 signal_name = ""
 scope = ""
 extension_script = ExtResource("1_of577")
+hidden = false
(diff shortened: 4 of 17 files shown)


wjt:
I'm trying a third version with the ability to `alias` blocks...

dbnicholson:
> > 1. Keep an unexported `spawn_frequency` that wraps `spawn_period` with getters and setters.
> 
> Actually it has to be exported; otherwise values in existing `.tscn` files are not set. I learned about the [`@export_storage`](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_exports.html#export-storage) annotation, added in Godot 4.3, which does what we want here. However that does mean bumping the minimum supported Godot version. And also it means that both properties will be saved in scenes – not only existing scenes which instantiated SimpleSpawner and set the old property, but also those that set the new name.

Huh, I thought I tried to do that in a test project and it worked. Testing again, I see it doesn't. Now I'm not nearly as ambitious about handling this.

wjt:
I tried adding an `aliases` property to `BlockDefinition`. After a bit of plumbing, it works correctly. But we still need to add the backwards-compat alias to the property itself (or else the `spawn frequency` property block, which is used in the sample scene, stops working); and we also need a way to alias block parameters, because the current `set spawn frequency` block is defined thusly:

`set spawn frequency to {new_frequency: FLOAT}`

so if I change this to:

`set spawn period to {new_period: FLOAT}`

then even with an alias so the block is still found, its parameter gets lost.

I really don't have the appetite for adding named parameter aliases.

Anyway, I'll push my block alias branch for posterity but I'm not going to work on it further. I think we should either go with this one, or (my preference) #272.

dbnicholson:
> I really don't have the appetite for adding named parameter aliases.

Indeed, let's just carry on with this. Ultimately I think all of this needs to be handled by adding some kind of scene migration. Unfortunately, I don't know how you'd do that exactly.

PR #286, opened 2024-10-23
Fix script popup

manuq:
Here is how it looks now with dark and light themes:
Captura desde 2024-10-23 11-01-24
Captura desde 2024-10-23 11-00-34

### DIFF
--- a/addons/block_code/ui/script_window/script_window.gd
+++ b/addons/block_code/ui/script_window/script_window.gd
@@ -18,8 +18,16 @@ func _apply_editor_syntax_highlighter():
 			break
 
 
+## Undoes the effect of the CodeEdit being read-only
+func _remove_font_color_alpha_clamp():
+	var font_readonly_color = script_label.get_theme_color("font_readonly_color")
+	font_readonly_color.a = 1
+	script_label.add_theme_color_override("font_readonly_color", font_readonly_color)
+
+
 func _ready():
 	_apply_editor_syntax_highlighter()
+	_remove_font_color_alpha_clamp()
 	script_label.text = script_content.replace("\t", "    ")
 
 
--- a/addons/block_code/ui/script_window/script_window.gd
+++ b/addons/block_code/ui/script_window/script_window.gd
@@ -3,7 +3,7 @@ extends Window
 
 var script_content: String = ""
 
-@onready var script_label: TextEdit = $Margin/VBox/ColorRect/Scroll/Margin/Code
+@onready var script_label: TextEdit = %Code
 
 
 func _ready():
--- a/addons/block_code/ui/script_window/script_window.gd
+++ b/addons/block_code/ui/script_window/script_window.gd
@@ -7,7 +7,6 @@ var script_content: String = ""
 
 
 func _ready():
-	popup_centered()
 	script_label.text = script_content.replace("\t", "    ")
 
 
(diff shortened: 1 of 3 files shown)


manuq:
Update, changing from TextEdit to CodeEdit displays a better / more readable background:
Captura desde 2024-10-23 11-21-28
Captura desde 2024-10-23 11-22-07

wjt:
This is a nice improvement though the code is still not always readable. For example, strings on a light theme are totally illegible:

Screenshot from 2024-10-23 16-30-36

I had the idea that it might be possible to reuse the syntax highlighting theme from the editor itself. It kind of is, with the caveat that it only works if you have at least one script open in the editor somewhere (even in a background tab). This is the result:

Screenshot from 2024-10-23 16-31-39

The colours are still not quite the same as in the real editor:

Screenshot from 2024-10-23 16-34-12

This is because we set the `editable` property to false. Spelunking through the TextEdit code, if I'm understanding correctly the problem is probably here:

https://github.com/wjt/godot/blob/f4942b735097c99b5e4882ec0b01b5a8df01a071/scene/gui/text_edit.cpp#L1307-L1308

At any given point, the text color's alpha channel is clamped to the `font_readonly_color` alpha, which happens to be 0.65.

I'll push another patch to fix that which gives this result which I'm quite satisfied with:

Screenshot from 2024-10-23 17-01-55
Screenshot from 2024-10-23 17-01-40

wjt (approved):
I also pushed a fixup for one of your commits.

WDYT?

wjt:
BTW are you resizing the window when you take these screenshots?

If not… I wonder why the window is so narrow on my system. The window allegedly has a default size of 750×750 pixels which is not the case either on your system or on mine.

addons/block_code/ui/script_window/script_window.gd
```diff
@@ -6,7 +6,20 @@ var script_content: String = ""
...
 
+## Attempts to match the syntax highlighting for some open script in the
+## editor, which is more likely to be appropriate for the editor theme
+## than our hardcoded default
+func _apply_editor_syntax_highlighter():
+	var script_editor: ScriptEditor = EditorInterface.get_script_editor()
+	for x in script_editor.get_open_script_editors():
+		var control: Control = x.get_base_editor()
+		if control is TextEdit:
+			script_label.syntax_highlighter = control.syntax_highlighter
```
manuq, line 17: Awesome!

addons/block_code/ui/script_window/script_window.gd
```diff
@@ -18,8 +18,16 @@ func _apply_editor_syntax_highlighter():
 			break
 
 
+## Undoes the effect of the CodeEdit being read-only
+func _remove_font_color_alpha_clamp():
+	var font_readonly_color = script_label.get_theme_color("font_readonly_color")
+	font_readonly_color.a = 1
+	script_label.add_theme_color_override("font_readonly_color", font_readonly_color)
```
manuq, line 25: Good finding, I didn't know the read-only was doing that. Thanks for digging the sourcecode.

manuq:
> BTW are you resizing the window when you take these screenshots?
> 
> If not… I wonder why the window is so narrow on my system. The window allegedly has a default size of 750×750 pixels which is not the case either on your system or on mine.

No I didn't, instead I added a lot of blocks to make it scroll. Hmm.

Issue #271, opened 2024-10-13
expose variable type on hover


cassidyjames:
### What Happened

We fought with what we thought was a bug but we were trying to snap a string into a float spot

### How to Reproduce

1. Create a string variable
2. Create a comparison block
3. Try to drag the string into the comparison, forgetting it's a string
4. Be sad

DoomTas3r:
I thought it would be like the preview block. The extra backgroud shape variants can also be used to express different types

cassidyjames:
Having it on hover is better than not having it at all. 😉 But yeah, something more visual could be helpful. I like the idea of distinct shapes for distinct types, but we'd want to make sure that's consistent across all the blocks.

I do still think being explicit with a tooltip (in addition to anything visual) would be beneficial and in line with our design of natural language plus helpful/technical tooltips elsewhere, though. Ultimately we're designing primarily for laptop/desktop users for our learning programs.

Maybe we add a tooltip as a first pass, and then do a more thorough follow-up with distinct shapes across all block types?

manuq:
@cassidyjames @DoomTas3r sounds good. We already have a tooltip, so it's a matter of adding a new line with the type. For example, for a Node2D under "Transform | Position", the "position" value block should say:

> The position property
>
> Type: Vector2

Long term, we have considered different shapes in the past, but there are so many that it will become a bit weird. I think an icon would be better.

Also at one point we would like to separate Vector2 in the 2 components, so there is one less type.

PR #287, opened 2024-10-23
Include type in tooltip

DoomTas3r:
The variant name as a string is appended to the description. If null, then no type is shown

Helping with https://github.com/endlessm/godot-block-coding/issues/271#issue-2583897016
image
image
image
image
image
image

### DIFF
--- a/addons/block_code/ui/blocks/block/block.gd
+++ b/addons/block_code/ui/blocks/block/block.gd
@@ -214,7 +214,13 @@ func _to_string():
 
 
 func _get_tooltip(at_position: Vector2) -> String:
-	return definition.description if definition else ""
+	if not definition:
+		return ""
+
+	if definition.variant_type == Variant.Type.TYPE_NIL:
+		return definition.description
+
+	return "{description}\n\nType: [b]{type}[/b]".format({"description": definition.description, "type": type_string(definition.variant_type)})
 
 
 func _make_custom_tooltip(for_text) -> Control:


cassidyjames (approved):
I'll let someone else review the implementation/code itself, but I like this from a UX perspective. :) It keeps the slightly technical information in the tooltip where we're already exposing that sort of thing, and displays it consistently. 👍🏻

In the spirit of progressive disclosure, it can help learners (or mentors/instructors!) better understand how/why things work the way they do, without overloading the block UI itself.

addons/block_code/ui/blocks/block/block.gd
```diff
@@ -214,7 +214,40 @@ func _to_string():
...
+		elif definition.variant_type == Variant.Type.TYPE_SIGNAL:
+			variant_as_string = "Signal"
+		elif definition.variant_type == Variant.Type.TYPE_DICTIONARY:
+			variant_as_string = "Dictionary"
+		elif definition.variant_type == Variant.Type.TYPE_ARRAY:
+			variant_as_string = "Array"
+		else:
+			variant_as_string = "Undefined"
+
+		return definition.description + "\n\nType: [b]" + variant_as_string + "[/b]"
```
wjt, line 249: Can you instead do this:

```suggestion
		return definition.description + "\n\nType: [b]" + type_string(definition.variant_type)
```

wjt: except perhaps not appending the type in the NIL case, as you already did. I'll edit my suggestion

wjt: Thanks for amending this. When I got back to my computer to test the branch I thought it would be clearer to use a format string, so I took the liberty of making that change in your patch.

wjt:
This is a really nice improvement. Thanks for thinking of it & submitting it!

PR #290, opened 2024-10-24
Handle SimpleSpawner compatibility

dbnicholson:
This adds handling for the SimpleSpawner compatibility breakage in #272. While I don't expect it to significantly affect our users, I took it as a chance to investigate how these issues could be handled. It shows how to handle a renamed property and a renamed block. I added everyone as reviewers to see what I came up with.

The added test validates the migration, but you can also see it for yourself by loading the `res://tests/data/simple_spawner_compat.tscn` scene in the editor. It should show that the scene is unsaved. After saving you'll see the names in the scene file have been updated appropriately.

https://phabricator.endlessm.com/T35547

### DIFF
--- a/addons/block_code/serialization/block_serialization.gd
+++ b/addons/block_code/serialization/block_serialization.gd
@@ -17,7 +17,9 @@ func _init(p_name: StringName = &"", p_children: Array[BlockSerialization] = [],
 
 
 # Block name and arguments backwards compatibility handling.
-const _renamed_blocks: Dictionary = {}
+const _renamed_blocks: Dictionary = {
+	&"simplespawner_set_spawn_frequency": &"simplespawner_set_spawn_period",
+}
 
 
 func _set_name(value):
@@ -31,7 +33,12 @@ func _set_name(value):
 		name = value
 
 
-const _renamed_arguments: Dictionary = {}
+const _renamed_arguments: Dictionary = {
+	&"simplespawner_set_spawn_period":
+	{
+		"new_frequency": "new_period",
+	},
+}
 
 
 func _set_arguments(value):
--- a/addons/block_code/serialization/block_serialization.gd
+++ b/addons/block_code/serialization/block_serialization.gd
@@ -1,13 +1,64 @@
+@tool
 extends Resource
 
 const BlockSerialization = preload("res://addons/block_code/serialization/block_serialization.gd")
 
-@export var name: StringName
+@export var name: StringName:
+	set = _set_name
 @export var children: Array[BlockSerialization]
-@export var arguments: Dictionary  # String, ValueBlockSerialization
+@export var arguments: Dictionary:  # String, ValueBlockSerialization
+	set = _set_arguments
 
 
 func _init(p_name: StringName = &"", p_children: Array[BlockSerialization] = [], p_arguments: Dictionary = {}):
 	name = p_name
 	children = p_children
 	arguments = p_arguments
+
+
+# Block name and arguments backwards compatibility handling.
+const _renamed_blocks: Dictionary = {}
+
+
+func _set_name(value):
+	var new_name = _renamed_blocks.get(value)
+	if new_name:
+		print("Migrating block %s to new name %s" % [value, new_name])
+		name = new_name
+		if Engine.is_editor_hint():
+			EditorInterface.mark_scene_as_unsaved()
+	else:
+		name = value
+
+
+const _renamed_arguments: Dictionary = {}
+
+
+func _set_arguments(value):
+	if not value is Dictionary:
+		return
+
+	var renamed_args = _renamed_arguments.get(name)
+	if not renamed_args:
+		# Try with the new block name if it hasn't been migrated yet.
+		var new_block_name = _renamed_blocks.get(name)
+		if new_block_name:
+			renamed_args = _renamed_arguments.get(new_block_name)
+
+	if renamed_args:
+		var changed: bool = false
+		value = value.duplicate()
+		for old_arg in renamed_args.keys():
+			if not old_arg in value:
+				continue
+
+			var new_arg = renamed_args[old_arg]
+			print("Migrating block %s argument %s to new name %s" % [name, old_arg, new_arg])
+			value[new_arg] = value[old_arg]
+			value.erase(old_arg)
+			changed = true
+
+		if changed and Engine.is_editor_hint():
+			EditorInterface.mark_scene_as_unsaved()
+
+	arguments = value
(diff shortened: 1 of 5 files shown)


addons/block_code/serialization/block_serialization.gd
```diff
@@ -11,3 +13,20 @@ func _init(p_name: StringName = &"", p_children: Array[BlockSerialization] = [],
...
+
+func _set_name(value):
+	var new_name = _renamed_blocks.get(value)
+	if new_name:
+		print("Migrating block %s to new name %s" % [value, new_name])
+		name = new_name
+		if Engine.is_editor_hint():
+			EditorInterface.mark_scene_as_unsaved()
+	else:
+		name = value
```
dylanmccall, line 32: I'm happy with just this as a place to start. Two thoughts for the distant future:

- There's a danger that the old name is now unusable in this case, so we might want to couple it with a version number or something.
- This requires that we maintain a list of renamed blocks in the core plugin code. That's an issue for blocks added outside of the core plugin. Those aren't a thing at the moment, but a different (albeit fussier) approach could be a `renamed_from` field in block_definition.gd.

addons/block_code/simple_spawner/simple_spawner.gd
```diff
@@ -168,3 +168,40 @@ static func setup_custom_blocks():
...
+		{
+			# spawn_frequency was renamed to spawn_period
+			"name": "spawn_frequency",
+			"class_name": &"",
+			"type": TYPE_FLOAT,
+			"hint": PROPERTY_HINT_NONE,
+			"hint_string": "",
+			"usage": PROPERTY_USAGE_NONE,
+		},
+	]
```
dylanmccall, line 185: Ooh, interesting!

addons/block_code/simple_spawner/simple_spawner.gd
```diff
@@ -168,3 +168,40 @@ static func setup_custom_blocks():
...
+
+
+func _get(property: StringName) -> Variant:
+	match property:
+		"spawn_frequency":
+			return spawn_period
+		_:
+			return null
```
dylanmccall, line 193: I'm somewhat disturbed that returning null means "handle the property normally" (ergo a property can't just be null), but, indeed. This is fun!

tests/test_compatibility.gd
```diff
@@ -0,0 +1,114 @@
...
+		"simplespawner_get_spawn_frequency",
+		"simplespawner_set_spawn_frequency",
+	]
+
+	const new_block_names: Array[String] = [
+		"simplespawner_get_spawn_period",
+		"simplespawner_set_spawn_period",
+	]
+
+	var scene: PackedScene = load("res://tests/data/simple_spawner_compat.tscn")
```
dylanmccall, line 52: Ooh, it's marvellous to have an example of a test with a scene file! Thank you for writing tests :)

addons/block_code/serialization/block_serialization.gd
dbnicholson: Indeed, this only handles pure renames. If you changed the arguments or something like that, this wouldn't work. We do have a version number [in the higher level BlockScriptSerialization](https://github.com/endlessm/godot-block-coding/blob/2f66ee7f9489deace3becce1ba0478281bdaad81/addons/block_code/serialization/block_script_serialization.gd#L29), but it's not available here right now.

@wjt poked at adding aliases for blocks in #284. I didn't look closely at his approach there, but it was my initial suggestion :laughing:

wjt: Unfortunately I actually did change the argument name in the "set ~~frequency~~ period" block...

addons/block_code/serialization/block_serialization.gd
```diff
@@ -16,7 +16,9 @@ func _init(p_name: StringName = &"", p_children: Array[BlockSerialization] = [],
 
 
 # Block name backwards compatibility handling.
-const _renamed_blocks: Dictionary = {}
+const _renamed_blocks: Dictionary = {
+	&"simplespawner_set_spawn_frequency": &"simplespawner_set_spawn_period",
```
wjt, line 20: I guess one could extend this so that the values are not just a string but a tuple of:

- new block name
- map from old parameter name to new parameter name

dbnicholson: Yes, something like that. An issue here is that we don't control the order that Godot sets the property in. So, name might not be migrated in this case. Here's an untested thing I came up with:

```
diff --git a/addons/block_code/serialization/block_serialization.gd b/addons/block_code/serialization/block_serialization.gd
index 4332b73..46ffd4b 100644
--- a/addons/block_code/serialization/block_serialization.gd
+++ b/addons/block_code/serialization/block_serialization.gd
@@ -6,7 +6,8 @@ const BlockSerialization = preload("res://addons/block_code/serialization/block_
 @export var name: StringName:
 	set = _set_name
 @export var children: Array[BlockSerialization]
-@export var arguments: Dictionary  # String, ValueBlockSerialization
+@export var arguments: Dictionary:  # String, ValueBlockSerialization
+	set = _set_arguments
 
 
 func _init(p_name: StringName = &"", p_children: Array[BlockSerialization] = [], p_arguments: Dictionary = {}):
@@ -30,3 +31,35 @@ func _set_name(value):
 			EditorInterface.mark_scene_as_unsaved()
 	else:
 		name = value
+
+
+const _renamed_arguments: Dictionary = {
+	&"simplespawner_set_spawn_period": {
+		"new_frequency": "new_period",
+	},
+}
+
+
+func _set_arguments(value):
+	if not value is Dictionary:
+		return
+
+	var renamed_args = _renamed_arguments.get(name)
+	if not renamed_args:
+		# Try with the new block name if it hasn't been migrated yet.
+		var new_block_name = _renamed_blocks.get(name)
+		if new_block_name:
+			renamed_args = _renamed_arguments.get(new_block_name)
+
+	if renamed_args:
+		value = value.duplicate()
+		for old_arg in renamed_args.keys():
+			if not old_arg in value:
+				continue
+
+			var new_arg = renamed_args[old_arg]
+			print("Migrating block %s argument %s to new name %s" % [name, old_arg, new_arg])
+			value[new_arg] = value[old_arg]
+			value.erase(old_arg)
+
+	arguments = value
```

addons/block_code/simple_spawner/simple_spawner.gd
dbnicholson: Indeed, `null` as sentinel and value is always fun.

dbnicholson:
I added block argument rename handling as described above. The required `new_frequency` to `new_period` migration wasn't happening before, but now it is. This also highlights that I hate that `BlockSerialization` and `ValueBlockSerialization` are entirely duplicated classes.

manuq (approved):
It is fantastic to see this migration happening, with unit test even!

PR #291, opened 2024-10-25
Add tooltip code details

dylanmccall:
Just idly playing with tooltips here. Let me know what you think?

Some nice examples:

Screenshot From 2024-10-24 21-01-40

Screenshot From 2024-10-24 21-01-55

Screenshot From 2024-10-24 21-08-39

Screenshot From 2024-10-24 21-02-15

Screenshot From 2024-10-24 21-02-42

Some more suspect examples:

Screenshot From 2024-10-24 21-03-12

Screenshot From 2024-10-24 21-04-32

I was hoping I could improve the formatting here with some divider lines, which would especially help when we have both parameters and an output value, but Godot's bbcode implementation doesn't seem to support `[hr]`, so we'd have to do a whole other thing with the custom tooltip scene (and that whole thing should really involve adding syntax highlighting anyway).


starnight (approved):
Wow!  This looks amazing!

dbnicholson (approved):
Looks pretty sweet to me! @wjt you poked at tooltips and bbcode before. WDYT?

DoomTas3r:
Another interesting example:
capture

wjt:
Personally I think this is too much information. I would want to get e.g. Justin's opinion as an educator.

DoomTas3r:
Maybe the tooltip could include the minimum information and an option in the context menu would have the full summary
capture

manuq:
Converting to Draft until we get feedback from Justin or any other educator.

PR #292, opened 2024-10-25
SimpleSpawner: Add "rotate with parent" property

manuq:
Add yet another setting to the SimpleSpawner. This is to rotate the spawned scene according to the SimpleSpawner parent node. Useful for changing the direction of bullets spawned from a rotated object.


wjt:
In what situation do you not want this to be `true`? E.g. for the spaceship runner, it would have no effect either way since the obstacle/coin spawner doesn't rotate.

manuq:
> In what situation do you not want this to be `true`? E.g. for the spaceship runner, it would have no effect either way since the obstacle/coin spawner doesn't rotate.

I was only being extra careful with backwards compatibility. But yeah having `true` as default doesn't break either the existing demo or the example scene.

manuq:
Converting this to draft because I realize that I can use the global rotation, rather than the parent's rotation. And to address feedback from @dylanmccall over Slack.

manuq:
The alternative to this is: https://github.com/endlessm/godot-block-coding/pull/294

dylanmccall:
I'm closing this in favour of #294, which has been merged.

Issue #267, opened 2024-10-11
Round Block-Inserting Suggestion


5YSTM:
### What Happened

is:open label:enhancement Allow "Round" blocks to be placed in any round slot.

Ex: [Play sound "(variable value used as sound name)"] or [Change rotation degrees by "(Ui _Up pressed)"]

The "True" or "False" output, being replaced with a 1 or 0 output instead. Also maybe add small symbols on each block type and allow us to right click a block to delete it, or duplicate it. (It would just help coding feel more natural to me since I often use scratch) 

### How to Reproduce

Just a suggestion

### Block Coding Plugin Version

v0.7.1

### Godot Engine Version

v4.3

### Operating System

Windows 11

manuq:
Round blocks may have different types, so they can't be placed anywhere. For example a Vector3 round block can't be placed into a slot that expects a String block.

There are however some conversions happening and they should be improved.

DoomTas3r:
In my personal opinion, type conversion should be explicit, such as using a string to integer block, that way it's value can be accurately inferred. Then, types can also have unique visual indicators, such as shapes or icons, to avoid confusion about which type it is. In Scratch, if some code accidentally adds a string and an integer, the issue might be misinterpreted. Also, in GDScript, using the `+` operator between a string and an integer produces an error.

Issue #201, opened 2024-08-20
add duplicate function for easy use ($200 bounty)


FakeDivinity:
### What Happened

its a waste of time for doing duplication it manually

### How to Reproduce

add a function where i can haver the existing code block and duplicate it when release

### Block Coding Plugin Version

0.6.2

### Godot Engine Version

4.3

### Operating System

windows11

manuq:
Yes! This would be a great contribution.

manuq:
For someone wanting to implement this, here is a proposal:
- Display a dropdown menu when the user right-clicks on a block.
- Add options: Duplicate and Delete to the dropdown menu.
- Add shortcut Ctrl-D to Duplicate.

The Delete functionality is already there, but only available from the Del shortcut.

dsd:
Our partner org Endless Studios is offering a modest bounty for completion of this task
/bounty $50

manuq:
There is a PR that adds the duplicate functionality among other things https://github.com/endlessm/godot-block-coding/pull/293

I think that the other things should be discussed first. The duplicate is definitely wanted, so if they can be split into a separate PR (or the other stuff removed) I'll give a review.

Gyan-max:
Is this still available to be resolved? 
please let me know, if I can get assigned

manuq:
@dsd please award @urbit-pilled .

PR #293, opened 2024-10-25
Context menu, delete, duplicate, and pin function

DoomTas3r:
Adds a context menu with 4 functions: duplicate, summary, pin, and delete

Adds a duplicate function with `ctrl + d` as a shortcut as suggested by https://github.com/endlessm/godot-block-coding/issues/201#issuecomment-2422916623 and https://github.com/endlessm/godot-block-coding/issues/267#issue-2582213586. The offset is 100 pixels right, 50 pixels down. The snapped blocks and parameter values are also duplicated

The summary shows the tooltip of the block, intended to be replaced by a more detailed overview like https://github.com/endlessm/godot-block-coding/pull/291#issue-2613021742

Adds a pin function with `ctrl + p` as a shortcut. Pinned blocks can't be dragged. Snapped blocks can pin with their parent block. This feature doesn't need to be added because the undo feature can undo the effects of accidental drags or deletes, so this is just for convenience

`_confirm_delete` is its own function

Functions that check snapped blocks only check `Container` nodes because only they have `SnapPoint` nodes

When the drag drop area receives a right click, a new popup menu determines which function is called from the parent block
image
image

### DIFF
--- a/addons/block_code/ui/blocks/block/block.gd
+++ b/addons/block_code/ui/blocks/block/block.gd
@@ -41,7 +41,42 @@ signal modified
 ## Whether the block can be deleted by the Delete key.
 var can_delete: bool = true
 
+# FIXME: Variable pinned should be saved with the scene
+## Whether the block is pinned
+var pinned: bool:
+	set(value):
+		if not can_delete:
+			return
+
+		pinned = value
+
+		if pinned:
+			block_pinned_container = Container.new()
+			block_pinned_container.mouse_filter = Control.MOUSE_FILTER_IGNORE
+
+			var block_pinned_panel := Panel.new()
+			block_pinned_panel.custom_minimum_size = Vector2(16, 16)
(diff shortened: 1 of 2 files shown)


manuq:
@DoomTas3r thanks! There is a bounty for the duplicate functionality. Please follow the steps in https://github.com/endlessm/godot-block-coding/issues/201#issuecomment-2433252534 if you would like to attempt it.

DoomTas3r:
Thanks, but I will not accept or claim the bounty

manuq:
@DoomTas3r I see that you are trying to add pin functionality to this PR as well, which was never discussed. My suggestion is to try contributing smaller chunks, because they are more likely to be merged. For example I would start with a context menu and just the current Delete functionality that is already there (as Del key shortcut).

manuq:
@DoomTas3r the Duplicate and Delete has been merged. If you still want to discuss pin functionality, please open a issue. Closing this for now!

PR #294, opened 2024-10-28
Simple spawner rotation alt

manuq:
This is an alternative to https://github.com/endlessm/godot-block-coding/pull/292

Instead of adding yet another property, make the "rotate with spawner" the default. Also remove the "spawn parent" property to leave only the SCENE behavior (as a separate commit).

### DIFF
--- a/addons/block_code/simple_spawner/simple_spawner.gd
+++ b/addons/block_code/simple_spawner/simple_spawner.gd
@@ -1,6 +1,12 @@
 @tool
 class_name SimpleSpawner
 extends Node2D
+## SimpleSpawner node.
+##
+## The scene being spawned is rotated according to this node's global rotation:
+## - If the spawned scene is a RigidBody2D, the linear velocity and constant forces
+##   are rotated according to the SimpleSpawner node global rotation.
+## - If the spawned scene is a Node2D, the rotation is copied from the SimpleSpawner node.
 
 const BlockDefinition = preload("res://addons/block_code/code_generation/block_definition.gd")
 const BlocksCatalog = preload("res://addons/block_code/code_generation/blocks_catalog.gd")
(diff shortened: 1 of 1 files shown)


addons/block_code/simple_spawner/simple_spawner.gd
```diff
@@ -99,12 +100,15 @@ func spawn_once():
...
-			get_tree().current_scene.add_child(spawned)
-			spawned.position = global_position
+	# Rotate the spawned scene according to the SimpleSpawner:
+	if spawned is RigidBody2D:
+		spawned.linear_velocity = spawned.linear_velocity.rotated(global_rotation)
+		spawned.constant_force = spawned.constant_force.rotated(global_rotation)
+	elif spawned is Node2D:
+		spawned.rotate(global_rotation)
+	# Add the spawned instance to the current scene:
+	get_tree().current_scene.add_child(spawned)
```
dylanmccall, line 110: Sorry to be jumping in about unrelated things (it's just we never really had the chance to review this properly :b), but _while we're here_, is there a good reason this is dumping the spawned node in the root of the scene instead of doing `add_sibling(spawned)`? I can see how it was easier to express in the options list we had before, but at this point it's rather surprising behaviour.

manuq: Sorry, I didn't see your comment until now @dylanmccall.

I just tried the "siblings" alternative: https://github.com/endlessm/godot-block-coding/tree/simple-spawner-rotation-alt2

But for the space ship demo, it makes the blasters rotate with the ship:
[Grabación de pantalla desde 2024-10-30 10-58-13.webm](https://github.com/user-attachments/assets/a97b0f06-a2e6-4b66-8341-28413343c4f1)


This is because the SimpleSpawner is a child of the ship that's being rotated with keystrokes.

dylanmccall: Oooh right, that makes sense :) Thanks for reminding me!

dylanmccall (approved):
I like the feel of this a lot :) It makes sense to me that SimpleSpawner is just one of those oddly-shaped building blocks that lets you do cool things. I asked for a bit of feedback from Justin as well, to be sure.

manuq:
> I asked for a bit of feedback from Justin as well, to be sure.

Should we just go with this? I think we already gave enough time for early feedback.

dylanmccall:
Yeah, let's see what we hear back after the next release.

Issue #152, opened 2024-07-22
Add place for Russian localisation


Lyumih:
### What Happened

Greetings!
I found a video about your plugin in one of the Godot groups. I really liked it.

There is a proposal to add a place for localizing blocks (Russian and other languages).
This ticket is linked to https://github.com/endlessm/godot-block-coding/issues/146.

The community, or myself in particular, could help with the translation if such an opportunity arises.

Thank you!

### How to Reproduce

none

manuq:
@Lyumih thank you for the interest! We are under some architectural changes. After that is done, we'll be glad to receive localization contributions in any language. Let's leave these tickets open so we don't forget. I will let you know!

wjt:
Hi @Lyumih – it is now possible to translate the block coding plugin, though the translations will not be shown until Godot 4.4 due to limitations of 4.3 and older. See the README file for details.

Issue #146, opened 2024-07-20
Any localization of each text on the blocks? 


JustusPan:
### What Happened

How could I help to add a Chinese version for this tool?

### How to Reproduce

none

manuq:
@JustusPan  thank you for the interest! We are under some architectural changes. After that is done, we'll be glad to receive localization contributions in any language. Let's leave these tickets open so we don't forget. I will let you know!

wjt:
Hi @JustusPan  – it is now possible to translate the block coding plugin, though the translations will not be shown until Godot 4.4 due to limitations of 4.3 and older. See the README file for details.

PR #295, opened 2024-10-29
Translation infrastructure

dbnicholson:
This is a WIP branch I had to provide translations for the plugin. It would need some work regardless, but it turns out that [editor translation plugins don't work](https://github.com/godotengine/godot/issues/43553), so it's a moot point. Just putting it here for reference and in the unlikely scenario that Godot does start supporting editor plugin translations.

This has now been updated to use [TranslationDomain](https://docs.godotengine.org/en/latest/classes/class_translationdomain.html), which was added in Godot 4.4-dev3. If you run [4.4-dev3](https://godotengine.org/article/dev-snapshot-godot-4-4-dev-3/) and put some translations in the PO files, you'll see them in the editor.

Fixes: #146
Fixes: #152

https://phabricator.endlessm.com/T35705

Here's a tool script I was using to test that ultimately proved to me that translations weren't happening with `TranslationServer`:

```
@tool
extends EditorScript

func _run():
	var locale = TranslationServer.get_locale()
	var tool_locale = TranslationServer.get_tool_locale()
	print("locale: %s" % locale)
	print("tool locale: %s" % tool_locale)
	#print("languages: %s" % TranslationServer.get_all_languages())
	print("locales: %s" % TranslationServer.get_loaded_locales())

	var tx = TranslationServer.get_translation_object(locale)
	print("translation: %s" % tx)
	if tx:
		print("translation locale: %s" % tx.locale)
		print("translation message list: %s" % tx.get_translated_message_list())
	
	var tool_tx = TranslationServer.get_translation_object(tool_locale)
	print("tool translation: %s" % tool_tx)
	if tool_tx:
		print("tool translation locale: %s" % tool_tx.locale)
		print("tool translation message list: %s" % tool_tx.get_translated_message_list())
	
	print(TranslationServer.translate("Add the node into the group"))
```


dbnicholson:
Note that if you want to set the tool locale, you have to use `LANG`/`LC_*` environment variables or however you set the locale for a process. The Godot `--language` option only sets the runtime locale, not the editor locale.

dbnicholson:
Turns out you _can_ do this in 4.4 using the new [TranslationDomain](https://docs.godotengine.org/en/latest/classes/class_translationdomain.html) support. I added a couple commits to do that and it seems to work. Still needs cleanup, though. I'm going to reopen and mark as a draft.

dbnicholson:
This is really ready for review now. I tested with 4.3 and [4.4-dev3](https://godotengine.org/article/dev-snapshot-godot-4-4-dev-3/). The editor translations only show up in 4.4, but I added some bogus translations and they do work there.

addons/block_code/inspector_plugin/block_script_inspector.gd
```diff
@@ -11,7 +16,7 @@ func _parse_begin(object):
 	var block_code := object as BlockCode
 
 	var button := Button.new()
-	button.text = "Open Block Script"
+	button.text = tr("Open Block Script")
```
wjt, line 19: Should this be `TxUtils.translate("Open Block Script")`? If not, why not?

addons/block_code/ui/blocks/block/block.gd
```diff
@@ -217,10 +217,11 @@ func _get_tooltip(at_position: Vector2) -> String:
 	if not definition:
 		return ""
 
+	var description_tx := tr(definition.description)
 	if definition.variant_type == Variant.Type.TYPE_NIL:
-		return definition.description
+		return description_tx
 
-	return "{description}\n\nType: [b]{type}[/b]".format({"description": definition.description, "type": type_string(definition.variant_type)})
+	return "{description}\n\nType: [b]{type}[/b]".format({"description": description_tx, "type": type_string(definition.variant_type)})
```
wjt, line 224: We'll want to translate the string “Type: ” in here. Perhaps this should be broken up a little...

addons/block_code/simple_nodes/simple_ending/simple_ending.gd
```diff
@@ -21,12 +21,13 @@ func game_over(result: String):
...
 		_:
-			text = ""
+			reset()
 			push_warning('Unrecognized game result "%s"' % result)
 
 
 func reset():
-	text = ""
+	# Workaround for POT generation extracting empty string.
+	text = str("")
```
wjt, line 30: Weird, so it extracts constant strings for translation even if not tagged with `tr()`?

addons/block_code/block_code_plugin.gd
```diff
@@ -56,6 +56,7 @@ func _enter_tree():
 
 	# Custom Project->Tools menu items.
 	add_tool_menu_item(tr("Regenerate %s POT file") % "BlockCode", TxUtils.regenerate_pot_file)
+	add_tool_menu_item(tr("Update %s translated files") % "BlockCode", TxUtils.update_pot_files)
```
wjt, line 59: I wonder how feasible it would be to have a CI job that runs a (real, not headless) editor and then executes these...

addons/block_code/locale/pt.po
```diff
@@ -0,0 +1,853 @@
...
+# res://addons/block_code/ui/tooltip/tooltip.tscn
+#
+# Automatically generated, 2024.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: Block Coding Plugin\n"
+"Last-Translator: Automatically generated\n"
+"Language-Team: none\n"
+"Language: pt\n"
```
wjt, line 166: You could ask a friendly Portuguese speaker to confirm but my understanding is that pt_BR and pt are different enough to warrant different translations.

wjt (commented):
Very nice! Obviously I'm offended that there is no placeholder `en_GB.po` file.

addons/block_code/inspector_plugin/block_script_inspector.gd
dbnicholson: This is a great question. The documentation for [Object.tr](https://docs.godotengine.org/en/latest/classes/class_object.html#class-object-method-tr) hasn't been updated since `TranslationDomain` was added, but essentially it's a wrapper around `translate`. Prior to 4.4, that would be [TranslationServer.translate](https://docs.godotengine.org/en/latest/classes/class_translationserver.html#class-translationserver-method-translate). In 4.4, that would be [TranslationDomain.translate](https://docs.godotengine.org/en/latest/classes/class_translationdomain.html#class-translationdomain-method-translate) after looking up the [Object's translation domain](https://docs.godotengine.org/en/latest/classes/class_object.html#class-object-method-get-translation-domain). There are a couple reasons not to call `translate` directly:

1. `tr` and `tr_n` respect [Object.can_translate_messages](https://docs.godotengine.org/en/latest/classes/class_object.html#class-object-method-can-translate-messages).
2. The GDScript POT extractor will extract constant strings passed to `tr` and `tr_n` as translatable strings. Now I'm noticing that `TxUtils.translate("constant string")` won't do that and I'll need to add a hack for that.

The only reason you'd want to call `translate` directly is if you're in a static context since you can't use `Object.tr` in that case.

wjt: OK, thanks for the clear explanation!

dbnicholson: I think I should probably document this. Often times I forget that after I've gone deep on something and it becomes clear to _me_, it's still likely unclear for someone else.

addons/block_code/ui/blocks/block/block.gd
dbnicholson: Ah, good catch. I didn't even notice it in there. I think I can wrap the whole thing in `tr` and it'll work out like shown [here](https://docs.godotengine.org/en/stable/tutorials/i18n/internationalizing_games.html#placeholders).

wjt: Yeah we can but it seems a little unkind (and error prone) to translators to make them preserve the format string. Also if we later wanted to lay the tooltip out differently, e.g. using some kind of grid to present attributes in a table, we would want a standalone translation of `Type:`.

Personally I would mark `Type:` for translation and interpolate that into the final string.

addons/block_code/simple_nodes/simple_ending/simple_ending.gd
dbnicholson: It's actually a little uglier than that now that I've read through [the GDScript translation parser](https://github.com/godotengine/godot/blob/master/modules/gdscript/editor/gdscript_translation_parser_plugin.cpp). The assumption is that anything that assigns to a [`text` property](https://github.com/godotengine/godot/blob/c6c464cf9ae56e8b68620af65125dd980d0e8122/modules/gdscript/editor/gdscript_translation_parser_plugin.cpp#L254) is a `Label` or similar, which should be automatically translated. In this case, it is a subclass of a `Label`, so that's a correct assumption. However, I don't think it should try to translate an empty string.

I wouldn't have cared except that `msgmerge` errored on a duplicate empty string msgid.

addons/block_code/block_code_plugin.gd
dbnicholson: Good questoin. The `TxUtils.update_pot_files` one is independent of the editor, so you could write a headless `SceneTree` script that we could run from CI. I almost started doing it but decided to wait for now since `TxUtils.regenerate_pot_file` is very much dependent on the editor and I had no idea if that could be scripted.

addons/block_code/locale/pt.po
dbnicholson: You're probably right. What I really want to avoid can be demonstrated with the `es_AR` locale. Since the editor has a generic `es` locale and the `es_AR` locale, it picks `es_AR` as the locale for any Spanish locale like `es_ES` or `es_MX`. It seems the only time `TranslationServer.get_tool_locale()` returns `es` is when I run with the bogus locale `LANG=es.utf-8`. So, I think that if we have the `es_AR` and `pt_BR`, then for all practical purposes the `es` and `pt` translations are unused. This seems like a bug in Godot - it should prefer the country-less locale if there's not a locale matching your country.

wjt: That does seem like a bug!

If it picked `es_AR` for any latin-american Spanish locale but `es` for spanish-of-spain then that might make sense as a special-case because (IIUC) latin american dialects of spanish are more similar to one another than to spanish-of-spain...

dbnicholson: To demonstrate, save the following script:

```
@tool
extends EditorScript

func _run():
	var lang = OS.get_environment("LANG")
	var locale = TranslationServer.get_locale()
	var tool_locale = TranslationServer.get_tool_locale()
	print("LANG: %s" % lang)
	print("locale: %s" % locale)
	print("tool locale: %s" % tool_locale)
```

You can run this in the Script editor from its File menu or via the Ctrl+Shift+X shortcut. It's the closest thing to a REPL. Try that launching Godot with various `LANG` settings. You can also find out that the command line `--language` option only affects the game locale, not the tool locale.

dbnicholson: For example:
```
LANG: es_ES.utf-8
locale: es_AR
tool locale: es_AR
```

dbnicholson: On the other hand, since there's only a country-less `de` translation:
```
LANG: de_CH.utf-8
locale: de
tool locale: de
```
Or if there's no translation, it just returns the language:
```
LANG: en_GB.utf-8
locale: en
tool locale: en
```

addons/block_code/ui/blocks/block/block.gd
dbnicholson: Yeah, that makes sense. I'll fix that up.

addons/block_code/locale/pt.po
dbnicholson: Ah, the bug appears to be introduced [here](https://github.com/godotengine/godot/commit/945645410905e8b23228965e29c2be274cb12645#diff-985565fb1f983acb2d87dbe8931c929d0f46fdecb1d99a31e495b151c0f52186R380) with the `>` change to `>=`. Both `es` and `es_AR` have a score of 1. `es` comes first in the list and previously would have remained as the best matched locale. The necessary part for https://github.com/godotengine/godot/issues/57125 is `score > 0` so that you don't use a locale that's totally unrelated.

dbnicholson: Looks like someone else hit this in https://github.com/godotengine/godot/issues/90677.

dbnicholson: I updated the commit message with information about this bug.

addons/block_code/inspector_plugin/block_script_inspector.gd
dbnicholson: I added some info in the README about this.

addons/block_code/simple_nodes/simple_ending/simple_ending.gd
wjt: Weird. I can see the appeal of not having to litter the code base with `tr()` at some level, but it seems error-prone.

addons/block_code/block_code_plugin.gd
wjt: One for another day I think.

wjt (approved):

PR #296, opened 2024-10-30
BlockCanvas: Implement drag & drop the node's property from Inspector

starnight:
Dragging the node's property from Inspector shows the object's type as "obj_property". So, allow the "obj_property" type in _can_drop_data() and handle it in _drop_data(). Then, implement getting the property's value as a Variable block with _drop_obj_property() in detail.

And, if the modifier CTRL key is pressed, then drop a setting the property's value block.

https://phabricator.endlessm.com/T35649

### DIFF
--- a/addons/block_code/ui/block_canvas/block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/block_canvas.gd
@@ -92,6 +92,11 @@ func _can_drop_data(at_position: Vector2, data: Variant) -> bool:
 
 
 func _drop_data(at_position: Vector2, data: Variant) -> void:
+	if data["type"] == "nodes":
+		_drop_node(at_position, data)
+
+
+func _drop_node(at_position: Vector2, data: Variant) -> void:
 	var abs_path: NodePath = data.get("nodes", []).pop_back()
 	if abs_path == null:
 		return
--- a/addons/block_code/ui/block_canvas/block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/block_canvas.gd
@@ -4,9 +4,11 @@ extends MarginContainer
 const ASTList = preload("res://addons/block_code/code_generation/ast_list.gd")
 const BlockAST = preload("res://addons/block_code/code_generation/block_ast.gd")
 const BlockCodePlugin = preload("res://addons/block_code/block_code_plugin.gd")
+const BlockDefinition = preload("res://addons/block_code/code_generation/block_definition.gd")
 const BlockTreeUtil = preload("res://addons/block_code/ui/block_tree_util.gd")
 const DragManager = preload("res://addons/block_code/drag_manager/drag_manager.gd")
 const ScriptGenerator = preload("res://addons/block_code/code_generation/script_generator.gd")
+const Types = preload("res://addons/block_code/types/types.gd")
 const Util = preload("res://addons/block_code/ui/util.gd")
 
 const EXTEND_MARGIN: float = 800
@@ -75,6 +77,10 @@ func _can_drop_data(at_position: Vector2, data: Variant) -> bool:
 	if typeof(data) != TYPE_DICTIONARY:
 		return false
 
+	# Allow dropping property block
+	if data.get("type", "") == "obj_property":
+		return true
+
 	var nodes: Array = data.get("nodes", [])
 	if nodes.size() != 1:
 		return false
@@ -94,6 +100,8 @@ func _can_drop_data(at_position: Vector2, data: Variant) -> bool:
 func _drop_data(at_position: Vector2, data: Variant) -> void:
 	if data["type"] == "nodes":
 		_drop_node(at_position, data)
+	elif data["type"] == "obj_property":
+		_drop_obj_property(at_position, data)
 
 
 func _drop_node(at_position: Vector2, data: Variant) -> void:
@@ -109,6 +117,30 @@ func _drop_node(at_position: Vector2, data: Variant) -> void:
 
 	var block = _context.block_script.instantiate_block_by_name(&"get_node")
 	block.set_parameter_values_on_ready({"path": node_path})
+	add_block(block, at_position)
+	reconnect_block.emit(block)
+
+
+func _drop_obj_property(at_position: Vector2, data: Variant) -> void:
+	var object_name = str(data["object"]).get_slice(":", 0)
+	var property_name = data["property"]
+	var property_value = data["value"]
+
+	# Prepare a block to get the property's value
+	var block_definition = (
+		BlockDefinition
+		. new(
+			&"%s_get_%s" % [object_name, property_name],
+			object_name,
+			"The %s property" % property_name,
+			"Variables",
+			Types.BlockType.VALUE,
+			typeof(property_value),
+			"%s" % property_name.capitalize().to_lower(),
+			"%s" % property_name,
+		)
+	)
+	var block = _context.block_script.instantiate_block(block_definition)
 	add_block(block, at_position)
 	reconnect_block.emit(block)
 
--- a/addons/block_code/ui/block_canvas/block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/block_canvas.gd
@@ -54,6 +54,8 @@ var zoom: float:
 	get:
 		return _window.scale.x
 
+var _modifier_ctrl := false
+
 signal reconnect_block(block: Block)
 signal add_block_code
 signal open_scene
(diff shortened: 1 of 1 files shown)


starnight:
Here is the demo of how to drag & drop the node's property from Inspector
[Screencast From 2024-10-30 18-27-59.webm](https://github.com/user-attachments/assets/160b0bbb-ab98-44b5-80fe-c72573464811)

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -92,6 +98,40 @@ func _can_drop_data(at_position: Vector2, data: Variant) -> bool:
...
+		return
+
+	# Prepare a block to get the property's value
+	var block_definition = (
+		BlockDefinition
+		. new(
+			&"%s_get_%s" % [obj_name, prop_name],
+			obj_name,
+			"The %s property" % prop_name,
+			"Physics | Velocity",  # TODO: Should get this Category dynamically
```
wjt, line 122: Do we need to set a valid category? The block doesn't get added to the picker. Is this just for the sake of its colour?

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -92,6 +98,40 @@ func _can_drop_data(at_position: Vector2, data: Variant) -> bool:
...
+	if obj_name == "" || prop_name == "" || val == null:
+		return
+
+	# Prepare a block to get the property's value
+	var block_definition = (
+		BlockDefinition
+		. new(
+			&"%s_get_%s" % [obj_name, prop_name],
+			obj_name,
+			"The %s property" % prop_name,
```
wjt, line 121: BTW a few weeks ago I looked into whether it was possible to look up the documentation for a property from a gdscript addon. Sadly, it's not.

addons/block_code/ui/block_canvas/block_canvas.gd
starnight: Yes, for the color.

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -92,6 +98,40 @@ func _can_drop_data(at_position: Vector2, data: Variant) -> bool:
...
+	var prop_name = data.get("property", "")
+	var val = data.get("value", null)
+	if obj_name == "" || prop_name == "" || val == null:
+		return
+
+	# Prepare a block to get the property's value
+	var block_definition = (
+		BlockDefinition
+		. new(
+			&"%s_get_%s" % [obj_name, prop_name],
```
wjt, line 119: How about a setter?

addons/block_code/ui/block_canvas/block_canvas.gd
manuq: Maybe use the Variables category?

addons/block_code/ui/block_canvas/block_canvas.gd
manuq: I loosely remember Orchestrator (or was Godot 3 visual scripting?) using a modifier key while dropping to generate a setter. I don't remember if Shift or Ctrl.

wjt: Maybe we could add the property get & setter to the Variables section of the picker when it has been dragged from the inspector?

addons/block_code/ui/block_canvas/block_canvas.gd
starnight: Hmm!!! "Variables" sounds a good category!

addons/block_code/ui/block_canvas/block_canvas.gd
starnight: > Maybe we could add the property get & setter to the Variables section of the picker when it has been dragged from the inspector?

This is worth a new feature ticket.

starnight: Both:
> I loosely remember Orchestrator (or was Godot 3 visual scripting?) using a modifier key while dropping to generate a setter. I don't remember if Shift or Ctrl.

and

> Maybe we could add the property get & setter to the Variables section of the picker when it has been dragged from the inspector?

acknowledge https://github.com/endlessm/godot-block-coding/pull/296#discussion_r1822518307!

starnight: After think twice, using a modifier key while dropping to generate a setter mentioned in https://github.com/endlessm/godot-block-coding/pull/296#discussion_r1822525086 is a good, simple and quick implementation for now.

starnight:
Here is the demo including both set and get blocks of the property dragged from Inspector. 
[Screencast From 2024-10-31 15-15-53.webm](https://github.com/user-attachments/assets/aa3cb742-bcfc-4c6b-91c9-3b6d91451d02)

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -92,6 +100,14 @@ func _can_drop_data(at_position: Vector2, data: Variant) -> bool:
 
 
 func _drop_data(at_position: Vector2, data: Variant) -> void:
+	var type = data.get("type", "")
+	if type == "nodes":
+		_drop_node(at_position, data)
+	elif type == "obj_property":
```
manuq, line 106: Nit: at this point we already know that data has a "type" key, because `_can_drop_data` returned `true`.
```suggestion
	if data["type"] == "nodes":
		_drop_node(at_position, data)
	elif data["type"] == "obj_property":
```

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -108,6 +124,53 @@ func _drop_data(at_position: Vector2, data: Variant) -> void:
 	reconnect_block.emit(block)
 
 
+func _drop_obj_property(at_position: Vector2, data: Variant) -> void:
+	var obj_name = str(data.get("object", "")).get_slice(":", 0)
+	var prop_name = data.get("property", "")
```
manuq, line 129: Nit: use descriptive variable names:
```suggestion
	var object_name = str(data.get("object", "")).get_slice(":", 0)
	var property_name = data.get("property", "")
```

Also, it is worth the extra safety? Can't you just use `data["object"]`, `data["property"]` here? In other words. is there any property that when dropped doesn't come with the "object" or "property" keys in the data?

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -108,6 +124,53 @@ func _drop_data(at_position: Vector2, data: Variant) -> void:
...
+	var val = data.get("value", null)
```
manuq, line 130: Same nits a above for the `val` variable here. It is not a great variable name, and I'm not sure if the extra safety is needed.

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -108,6 +124,53 @@ func _drop_data(at_position: Vector2, data: Variant) -> void:
...
+		block_definition = (
+			BlockDefinition
+			. new(
+				&"%s_set_%s" % [obj_name, prop_name],
+				obj_name,
+				"Set the %s property" % prop_name,
+				"Variables",
+				Types.BlockType.STATEMENT,
+				prop_type,
+				"set %s to {value: %s}" % [prop_name.to_lower(), type_string],
```
manuq, line 149: I was about to say that this is now how we show other properties previously in the catalog, but you got it right. Previously we shown them differently, and that has regressed, so I will open an issue.

manuq (changes requested):
Only a couple nits requested. I notice that the property names in the catalog have regressed, so I will open an issue and a separate PR for them. This follows what's in the catalog, so it's OK.

manuq:
@starnight this is the regression I was talking about: https://github.com/endlessm/godot-block-coding/pull/303 . It is explained in the issue. If you agree, please adjust this PR accordingly.

addons/block_code/ui/block_canvas/block_canvas.gd
starnight: Sure.  Fixed accordingly

addons/block_code/ui/block_canvas/block_canvas.gd
starnight: Have not noticed any property that doesn't come with the "object" or "property" keys in the data, yet.  Maybe, they can be accessed directly.

addons/block_code/ui/block_canvas/block_canvas.gd
starnight: Yap!  Fixed.

addons/block_code/ui/block_canvas/block_canvas.gd
starnight: Access the "object", "property" and "value" within the new commit directly.

manuq (approved):
Looks good now! Please note that the Ctrl shortcut to obtain a setter is a bit hidden.

Issue #166, opened 2024-07-26
Sound: Allow dragging and dropping the audio file from the Filesystem dock ($200 bounty)


cassidyjames:
### What Happened

I was following along with our own [Block Coding pong tutorial on YouTube around 2:54](https://youtu.be/WlUN7Zz0Djg?si=kic2i7PvVBJ9MCDf&t=174), ~and tried to replicate~ _apparently hallucinated_ Will's action of dragging and dropping a file onto the path. It didn't work, so I had to right-click and copy the path.

I feel like drag and drop is going to be extra important to our target demographic here.

### How to Reproduce

1. Add a "Load file" block to the canvas
2. Try to drag an audio resource to the path

### Block Coding Plugin Version

v0.6.1

### Godot Engine Version

v4.2.2.stable.flathub [15073afe3]

### Operating System

Fedora Silverblue 40

manuq:
@cassidyjames from the video it seems that @wnbaum is also copy/pasting the path, not doing drag and drop.

Besides, I wonder if this should be easier. Eg. all `ogg` files in the project could be blocks that can be dragged directly from the Picker (the blocks palette).

manuq:
The easiest implementation would be:
- Allow dragging files from the Filesystem dock to the Block Coding canvas
- When dropped, a value block is created of type String and value `res://..../` (the resource path)
- Then this block can be attached to the 1st slot of the current "load file () as sound ()".

A much better implementation but also a more complex one:
- Allow dragging sound files from the Filesystem dock to the Block Coding canvas.
- When dropped, the sound is preloaded at the top of the generated script using the resource path. An AudioStreamPlayer is instantiated automatically in the `_ready()` method using that resource. And a value block is created of type String to reference the AudioStreamPlayer instance by name. The name can be the last part of the path using [`String.get_file()`](https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-get-file).
- The "play the sound (name)" and "stop the sound (name)" blocks would accept that block, but now the "load file () as sound (name)" block is not needed anymore.

dsd:
Our partner org Endless Studios is offering a modest bounty for completion of this task
(mostly thinking of the easy implementation mentioned above - that could be a good starting point even if we later do the more complex one)
/bounty $40

manuq:
> looks like the easy version of this has been already been solved with #305

Yep, closing this issue then.

PR #298, opened 2024-11-01
Add switch_scene resource

vanessa-chang:
Add a new block that allows the user to switch to another scene via a specified scene file path.

https://phabricator.endlessm.com/T35706

### DIFF
--- a/addons/block_code/blocks/communication/switch_scene.tres
+++ b/addons/block_code/blocks/communication/switch_scene.tres
@@ -1,0 +1,18 @@
+[gd_resource type="Resource" load_steps=2 format=3 uid="uid://dnc2555wnobks"]
+
+[ext_resource type="Script" path="res://addons/block_code/code_generation/block_definition.gd" id="1_rnqd5"]
+
+[resource]
+script = ExtResource("1_rnqd5")
+name = &"switch_scene"
+target_node_class = ""
+description = "Stop playing the current scene, and switch to a different one. You might use this to switch to a new level."
+category = "Communication | Methods"
+type = 2
+variant_type = 0
+display_template = "switch the scene to {file_path: STRING}"
+code_template = "get_tree().change_scene_to_file({file_path})
+"
+defaults = {}
+signal_name = ""
+scope = ""


addons/block_code/blocks/communication/switch_scene.tres
```diff
@@ -0,0 +1,18 @@
+[gd_resource type="Resource" load_steps=2 format=3 uid="uid://dnc2555wnobks"]
+
+[ext_resource type="Script" path="res://addons/block_code/code_generation/block_definition.gd" id="1_rnqd5"]
+
+[resource]
+script = ExtResource("1_rnqd5")
+name = &"switch_scene"
+target_node_class = ""
+description = "Delete the current scene and switch to a new scene"
```
starnight, line 9: How about "Delete the current scene and switch to the target scene"?

wjt: "delete" seems the wrong word. You're not deleting anything, just switching.

addons/block_code/blocks/communication/switch_scene.tres
```diff
@@ -0,0 +1,18 @@
...
+category = "Communication | Methods"
+type = 2
+variant_type = 4
+display_template = "switch the scene to {file_path: STRING}"
+code_template = "get_tree().change_scene_to_file({file_path})
```
wjt, line 14: People more expert in Godot than me: is there some way we can refer to the other scene other than by file path?

addons/block_code/blocks/communication/switch_scene.tres
wjt: Perhaps something like:

> Stop playing the current scene, and switch to a different one. You might use this to switch to a new level.

starnight: Simply place the Godot document [change_scene_to_file](https://docs.godotengine.org/en/stable/classes/class_scenetree.html#class-scenetree-method-change-scene-to-file)

starnight: According to [change_scene_to_packed](https://docs.godotengine.org/en/stable/classes/class_scenetree.html#class-scenetree-method-change-scene-to-packed)'s description relating to `change_scene_to_file()`: `At the end of the frame, the formerly current scene, already removed from the tree, will be deleted (freed from memory) and then the new scene will be instantiated and added to the tree.`, the "delete" happens.  However, the "delete" might be less important for beginners.

wjt: I would expect that using "delete" would be actively misleading for beginners. This is using `delete` in the "free()" sense, but in normal use when you "delete" something, it is gone forever.

addons/block_code/blocks/communication/switch_scene.tres
wjt: What I'm imagining here is: can we make it possible to drag a `tscn` file onto the canvas and then have a block representing a different scene?

vanessa-chang: There is another way to do is to export a PackedScene variable [packed scene](https://docs.godotengine.org/en/stable/classes/class_packedscene.html). It will allows the user to assign the scene in the inspector. Should we update and use this approach instead? (Although I'm not the expert) 
During my research, I found three ways to switch scenes: [T35706](https://phabricator.endlessm.com/T35706). Hopefully, we can get more input from the experts!

wjt: I think it is right to do this in the block coding canvas (as you have done) not to go to the inspector. I was hoping we could do something similar to how we can drop nodes into the canvas to get a block that represents that node.

I did some brief experiments with this... If you have a node that exports a variable of type `PackedScene`, then assign a scene to it, it gets represented as (wrapped for legibility):

```
[ext_resource type="PackedScene"
              uid="uid://c2pcat3ejl74o"
              path="res://Scene2.tscn"
              id="2_g4qc4"]
```

I feel like it ought to be possible to define a value block that gets serialised in a way that refers to an external scene in this way. Then this `switch_scene` block would accept a parameter of type PackedScene. And for bonus points we would give it a dropdown of all scenes in the project, like we do for parameters of type node or group.

Unfortunately block parameter types are expressed as [Variant.Type](https://docs.godotengine.org/en/stable/classes/class_@globalscope.html#enum-globalscope-variant-type) enum members and a `PackedScene` is just an `OBJECT`. There's no analogue to `NODE_PATH`.

As well as the drag-and-drop idea, another reason it may be nice to store the reference to another scene in a way that Godot itself understands to be a reference to a file rather than just a string is that Godot knows how to update all such references if you rename the scene.

Anyway... What I'm suggesting is of course all way more work than adding a switch_scene block whose parameter is a string. And having the block at all would make a whole world of things possible which currently aren't. So all my suggestions here should probably be considered further work for another time…

addons/block_code/blocks/communication/switch_scene.tres
wjt: I stand by my suggestion above:

```suggestion
description = "Stop playing the current scene, and switch to a different one. You might use this to switch to a new level."
```

wjt (changes requested):

addons/block_code/blocks/communication/switch_scene.tres
manuq: > What I'm imagining here is: can we make it possible to drag a `tscn` file onto the canvas and then have a block representing a different scene? Or a dropdown list of all `tscn` files in the project, except the current one and scenes that are part of the plugin?

Yeah, in the same way as the more sophisticated alternative to drag&drop sounds that I described in https://github.com/endlessm/godot-block-coding/issues/166#issuecomment-2442518814

> Unfortunately block parameter types are expressed as [Variant.Type](https://docs.godotengine.org/en/stable/classes/class_@globalscope.html#enum-globalscope-variant-type) enum members and a `PackedScene` is just an `OBJECT`. There's no analogue to `NODE_PATH`.

I think that we should add a `class_name` field to the block definition, that will pair with the `variant_type` field for the case where the variant is of type `OBJECT`. This is the same Godot has for typed arrays:
- https://docs.godotengine.org/en/stable/classes/class_array.html#class-array-method-get-typed-builtin
- https://docs.godotengine.org/en/stable/classes/class_array.html#class-array-method-get-typed-class-name

Then when dragging from the Filesystem to the block canvas:
- If the resource is a sound, add a value block of variant type OBJECT and class name AudioStream.
- If the resource is a `tscn` or `scn`, add a value block of variant type OBJECT and class name PackedScene.

In the same way, the holes or slots can be enhaced to allow a specific class name when the variant type is OBJECT. Then we could do:

- "play sound" block: Has a slot that supports a value block of variant type OBJECT and class name AudioStream.
- "switch scene" block: Has a slot that supports a value block of variant type OBJECT and class name PackedScene.

dbnicholson: Technically, a [PackedScene](https://docs.godotengine.org/en/stable/classes/class_packedscene.html) is a `Resource`. [SceneTree](https://docs.godotengine.org/en/stable/classes/class_scenetree.html) (what you get from `get_tree()`) provides 2 methods for changing scenes - from a [file path](https://docs.godotengine.org/en/stable/classes/class_scenetree.html#class-scenetree-method-change-scene-to-file), and from a [PackedScene](https://docs.godotengine.org/en/stable/classes/class_scenetree.html#class-scenetree-method-change-scene-to-packed).

In code you might do it like this:

```
func change_scene(scene: Variant):
    if scene is String:
        get_tree().change_scene_to_file(scene)
    elif scene is PackedScene:
        get_tree().change_scene_to_packed(scene)
    else:
        push_error("Can only change scene from String or PackedScene")
```

With blocks we kinda need to be specific about what kind of `Variant` the parameter is, so we'd need to do one or the other. @manuq has some suggestions on handling that if we want to use `PackedScene`s.

I like the idea of loading a `PackedScene` with `change_scene_to_packed`, but then you'd have to tackle the whole part of creating a scene with a block using `load`. It would be very similar to the `get_node` block, though.

starnight: I think #305 helps `get_tree().change_scene_to_file(scene)` :)

dbnicholson:
Having said everything about using a `PackedScene`, I think it's best to address @wjt's comments and get this in using `change_scene_to_file` using a `String`. If we have time later, we can add support for loading a scene as a block and using that as the argument to the `switch_scene` block. I'm confident we can add a migration should do the work to handle `PackedScene`s.

manuq (approved):
It works!

PR #299, opened 2024-11-04
Run tests with Godot 4.4 prerelease

wjt:
This is potentially interesting because we have translation-related code
that behaves differently between 4.4 (which has some new API) and older
versions (which do not).


wjt:
The Godot 4.4 process segfaults when running tests – not too promising.

wjt:
Updated this with `-dev7` which was released late last year.

PR #300, opened 2024-11-04
Update Gut addon to 9.3.0

wjt:
My original hope was that a newer Gut would avoid 4.4.0-dev3 crashing in a branch where I tried to add that version to the CI matrix. On closer inspection it's actually crashing at the `--import` step, and this doesn't help at all, but now I've done this update (which mainly involved working around a command line argument handling regression) I thought I may as well separate out these patches.


.github/workflows/checks.yaml
```diff
@@ -42,4 +42,4 @@ jobs:
         godot --path . --headless --import
     - name: Run tests
       run: |
-        godot --path . --headless --script addons/gut/gut_cmdln.gd -gexit
+        godot --path . --headless -s addons/gut/gut_cmdln.gd -gexit
```
dbnicholson, line 45: This is also referenced in the [README](https://github.com/endlessm/godot-block-coding/blob/main/addons/block_code/README.md#testing). Might you update those instructions, too?

dbnicholson (changes requested):

.github/workflows/checks.yaml
wjt: Done – I also added a note that `--script` can't be used.

dbnicholson (approved):

Issue #302, opened 2024-11-04
The label in blocks for propeties has regressed


manuq:
### What Happened

The Inspector has 3 ways to show the property names and defaults to Capitalized:
![Captura desde 2024-11-04 16-37-27](https://github.com/user-attachments/assets/49dcfe87-89b7-46a1-94b8-7fa872e7b474)

This is a nice idea as they are more human readable. Previously we did the same for properties in the catalog. But it has regressed in commit c18ab3ff (my fault!) so now they look more like the "Raw" Inspector option:
![Captura desde 2024-11-04 16-36-18](https://github.com/user-attachments/assets/16e205d4-6e82-4c21-9c10-0d1985187abb)

Since we switched to lower-cased labels, I think we should have capitalized plus lower case on top. So the example above should read "set rotation degrees to...". 

### How to Reproduce

Self explained.

### Block Coding Plugin Version

main

### Godot Engine Version

v4.3

### Operating System

Endless OS

PR #303, opened 2024-11-04
Switch property labels back to human readable

manuq:
In commit c18ab3f we regressed the property names from capitalized to lower case. We actually want both: capitalize them and then lower case them.

Fixes https://github.com/endlessm/godot-block-coding/issues/302


dbnicholson (approved):
I think that makes sense after reading that [capitalize](https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-capitalize) is more like a conversion to a sentence.

manuq:
> I think that makes sense after reading that [capitalize](https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-capitalize) is more like a conversion to a sentence.

Yes, it's not like Python's capitalize which makes the first character upper case and the rest lower-case.

PR #305, opened 2024-11-05
BlockCanvas: Implement drag & drop the file from resource FileSystem

starnight:
Dragging the files from resource FileSystem shows the files' type as
"files". So, allow the "files" type in _can_drop_data() and handle it in
_drop_data(). Then, implement getting the files' resource paths as
Variable blocks with _drop_files() in detail.

https://phabricator.endlessm.com/T35712

### DIFF
--- a/addons/block_code/code_generation/blocks_catalog.gd
+++ b/addons/block_code/code_generation/blocks_catalog.gd
@@ -302,3 +302,20 @@ static func get_property_setter_block_definition(variable: VariableDefinition) -
 	var block_def := get_variable_setter_block_definition(variable)
 	block_def.description = "Set the %s property" % variable.var_name
 	return block_def
+
+
+static func get_resource_block_definition(file_path: String) -> BlockDefinition:
+	var block_def := BlockDefinition.new()
+	var file_name = file_path.get_file()
+
+	# Block Definition's name cannot work with '.'
+	block_def.name = &"get_resource_file_path"
+	block_def.description = "The full resource path of '%s'" % file_name
+	block_def.category = "Variables"
+	block_def.type = Types.BlockType.VALUE
+	block_def.variant_type = TYPE_STRING
+	block_def.display_template = "%s {const file_path: STRING}" % file_name
+	block_def.code_template = "{file_path}"
+	block_def.defaults = {"file_path": file_path}
+
+	return block_def
--- a/addons/block_code/serialization/block_script_serialization.gd
+++ b/addons/block_code/serialization/block_script_serialization.gd
@@ -72,7 +72,7 @@ func instantiate_block(block_definition: BlockDefinition) -> Block:
 
 
 func instantiate_block_by_name(block_name: String) -> Block:
-	var block_definition := get_block_definition(block_name)
+	var block_definition := get_block_definition(block_name, {})
 
 	if block_definition == null:
 		push_warning("Cannot find a block definition for %s" % block_name)
@@ -81,18 +81,25 @@ func instantiate_block_by_name(block_name: String) -> Block:
 	return instantiate_block(block_definition)
 
 
-func get_block_definition(block_name: String) -> BlockDefinition:
+func get_block_definition(block_name: String, arguments: Dictionary) -> BlockDefinition:
 	var split := block_name.split(":", true, 1)
 	var block_definition: BlockDefinition
 
 	if len(split) > 1:
-		return _get_parameter_block_definition(split[0], split[1])
+		block_definition = _get_parameter_block_definition(split[0], split[1])
+		if block_definition:
+			return block_definition
 
 	block_definition = _get_base_block_definition(block_name)
 	if block_definition != null:
 		return block_definition
 
 	block_definition = _get_obj_property_block_definition(block_name)
+	if block_definition != null:
+		return block_definition
+
+	var file_path = arguments.get("file_path", "")
+	block_definition = _get_resource_block_definition(block_name, file_path)
 	if block_definition != null:
 		return block_definition
 
@@ -182,6 +189,16 @@ func _get_parent_node_property_info(property_name: String) -> Dictionary:
 	return {}
 
 
+func _get_resource_block_definition(block_name: String, file_path: String) -> BlockDefinition:
+	if block_name != "get_resource_file_path":
+		return null
+
+	if file_path.is_empty() or not FileAccess.file_exists(file_path) or not ResourceLoader.exists(file_path):
+		return null
+
+	return BlocksCatalog.get_resource_block_definition(file_path)
+
+
 func _update_block_definitions():
 	_available_blocks.clear()
 	_available_blocks.append_array(_get_inherited_block_definitions())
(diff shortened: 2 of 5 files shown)


starnight:
This depends on #296.  So, set it as draft now.

starnight:
Here is the demo video: 

https://github.com/user-attachments/assets/5b241ae7-7739-4cae-a4cf-173155e474e8

addons/block_code/code_generation/blocks_catalog.gd
```diff
@@ -302,3 +302,18 @@ static func get_property_setter_block_definition(variable: VariableDefinition) -
 	var block_def := get_variable_setter_block_definition(variable)
 	block_def.description = "Set the %s property" % variable.var_name
 	return block_def
+
+
+static func get_resource_block_definition(file_name: String) -> BlockDefinition:
+	var block_def := BlockDefinition.new()
+
+	# Block Definition's name cannot work with '.'
+	block_def.name = &"get_resource_%s" % file_name.replace(".", "_")
```
manuq, line 311: I wonder what happens if the file name has a whitespace in it. Is `file_name` here already escaping it?

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -182,6 +189,17 @@ func _get_parent_node_property_info(property_name: String) -> Dictionary:
 	return {}
 
 
+func _get_resource_block_definition(block_name: String, file_path: String) -> BlockDefinition:
+	if not block_name.begins_with("get_resource_"):
+		return null
+
+	if file_path == "":
```
manuq, line 196: From [docs](https://docs.godotengine.org/en/stable/classes/class_string.html#string):
> [is_empty](https://docs.godotengine.org/en/stable/classes/class_string.html#class-string-method-is-empty) should be used to check for empty strings.

```suggestion
	if file_path.is_empty():
```

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -182,6 +189,17 @@ func _get_parent_node_property_info(property_name: String) -> Dictionary:
...
+		return null
+
+	var file_name = file_path.get_file()
```
manuq, line 199: Is there a chance that this string doesn't have a valid file? Eg. what would happen if the user removes the file from the filesystem? In that case what does `get_file()` return?

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -148,6 +154,32 @@ func _drop_obj_property(at_position: Vector2, data: Variant) -> void:
...
 
 
+func _drop_files(at_position: Vector2, data: Variant) -> void:
+	var resource_files = data["files"]
+	var next_position = at_position
+	const bias = 20
+
+	for file_path in resource_files:
+		# Prepare a Variable block getting the file's resource path.
+		var block_definition = _res_file_block_def(file_path)
```
manuq, line 164: The function name `_res_file_block_def` is not self-explanatory. Maybe `_get_block_definition_from_resource_file()`? Or at least `_get_blockdef_from_resource_file()`?

By the way, that function is using `BlocksCatalog.get_resource_block_definition` which is more clear. Maybe the helper function is not needed at all?

manuq (approved):
I'm leaving some nits. Good work!

addons/block_code/code_generation/blocks_catalog.gd
dbnicholson, line 311: I feel like this would be better as a fixed block definition that sets the path as an argument since the serialization will handle that just fine. In fact, this is pretty much exactly the same as 4cc4b2845ab95817b0ded577e4107894532887fa with `get_node`. The `BlockExtension` there is neat, but the only really important part is setting the parameter value. I don't think these kinds of dynamically named blocks are generally a good idea since they rely on the serializer to resolve them.

addons/block_code/ui/blocks/utilities/template_editor/template_editor.gd
```diff
@@ -137,5 +141,12 @@ func _append_output_parameter(parameter: Dictionary, id: int):
 	_container.add_child(parameter_output)
 
 
+func _append_const_parameter(parameter: Dictionary, id: int):
+	# const_parameter is a kind of in_parameter with default value, but never
+	# changes value.
+	var parameter_const := _append_input_parameter(parameter, id)
+	parameter_const.visible = false
```
dbnicholson, line 148: What does this do exactly? I want to make sure the extra syntax is really worth it since it seems like the regular input parameter would work fine.

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -182,6 +189,17 @@ func _get_parent_node_property_info(property_name: String) -> Dictionary:
 	return {}
 
 
+func _get_resource_block_definition(block_name: String, file_path: String) -> BlockDefinition:
+	if not block_name.begins_with("get_resource_"):
+		return null
```
dbnicholson, line 194: This doesn't appear to use the name embedded in the block name at all. What was the use of that?

addons/block_code/code_generation/blocks_catalog.gd
starnight: According to the discussion https://github.com/endlessm/godot-block-coding/pull/305#discussion_r1859424547, the block's name can be a fixed name, like `get_resource_file_path`

addons/block_code/ui/block_canvas/block_canvas.gd
starnight: This comes from the following implementation of loading audio files, such as "ogg".  However, it is not included in this PR.  To be clear code, yes, using `BlocksCatalog.get_resource_block_definition()` may be a better way now.

addons/block_code/ui/blocks/utilities/template_editor/template_editor.gd
starnight: This clears the in parameter's "visible" to avoid modify the value from users.

starnight: Also, if the in parameter shows up, it displays the full file path as a long text.

addons/block_code/serialization/block_script_serialization.gd
starnight: yap!

addons/block_code/serialization/block_script_serialization.gd
starnight: `get_file()` simply returns the file name part of the "string".  Fixed in the new commit which checks with `FileAccess.file_exists()` and `ResourceLoader.exists()`.

addons/block_code/ui/blocks/utilities/template_editor/template_editor.gd
starnight: * Without the input parameter in the display:
  Screenshot From 2024-11-27 17-10-07
* With the input parameter in the display:
  Screenshot From 2024-11-27 17-12-16

addons/block_code/code_generation/blocks_catalog.gd
starnight: Good point!  The block name should be fixed here.  I set it as "get_resource_file_path" in the fixed commit.

addons/block_code/serialization/block_script_serialization.gd
starnight: I set it as "get_resource_file_path" in the fixed commit.

PR #306, opened 2024-11-05
Initial Transifex configuratoin

dbnicholson:
This adds configuration for the tx client and a pointer to the project. Later we might make use of the full Transifex/GitHub integration[1].

1. https://help.transifex.com/en/articles/6265125-github-installation-and-configuration

https://phabricator.endlessm.com/T35709

### DIFF
--- a/.tx/config
+++ b/.tx/config
@@ -1,0 +1,8 @@
+[main]
+host = https://www.transifex.com
+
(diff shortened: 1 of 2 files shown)


manuq (approved):
All looks good to me as I remember the Transifex config file.

PR #312, opened 2024-11-06
Pull translations from Transifex

manuq:
Run `tx pull --use-git-timestamps`. This brings a Spanish translation made by myself, and a partial German translation that's being contributed.

### DIFF
--- a/addons/block_code/locale/es.po
+++ b/addons/block_code/locale/es.po
@@ -175,11 +175,11 @@ msgstr ""
 
 #: addons/block_code/block_code_plugin.gd
 msgid "Regenerate %s POT file"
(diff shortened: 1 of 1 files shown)


dbnicholson:
What happened to the German translation?

manuq:
Update: I removed the partial DE translation as it's changing things between `{}` like "group" or "STRING", so it won't work:
> "ist in Gruppe {Gruppe: ZEICHENKETTE}"

dbnicholson (approved):
LGTM! In the future, we might have Transifex post these PRs. For a human maintainer like you, I feel like you can just push these directly if you've reviewed them.

PR #315, opened 2024-11-07
Fix "Type::" double colon in block tooltip

wjt:
In commit 3e208d75ccb761d936b3126082407632dc09d7ec, the string "Type:" was extracted from the template string to be translated separately, including The colon. It is correct for the colon to be part of the translatable string, because (for example) in French there must be a non-breaking space between the word and the colon.

However, there was also a stray colon in the template string. Remove it.


dbnicholson (approved):
:facepalm:

PR #316, opened 2024-11-08
BlockScriptSerialization: Re-generate block definition for object property blocks

starnight:
The drag & drop object property's blocks disappear after save & re-open the Godot project. And, shows error:

core/variant/variant_utility.cpp:1092 - Cannot construct block from null block definition. res://addons/block_code/ui/block_canvas/block_canvas.gd:348 - Invalid call. Nonexistent function 'set_parameter_values_on_ready' in base 'Nil'.

It is because the object property blocks are not the predefined blocks in the catalog. So, Block Coding plugin cannot find the block definition from the catalog when places the object property blocks into the cavas via _block_to_ast_node() after re-open the project.

Therefore, introduce _get_obj_property_block_definition() generating object property's getter/setter block definition for get_block_definition(). Besides, it also needs the object property's value type. So, pass the AST node's arguments which is a Dictionary and may contain the "value" feild into get_block_definition(), too.

Fixes: 89beea94eefa ("BlockCanvas: Implement drag & drop the node's property from Inspector") https://phabricator.endlessm.com/T35649

### DIFF
--- a/addons/block_code/code_generation/blocks_catalog.gd
+++ b/addons/block_code/code_generation/blocks_catalog.gd
@@ -253,25 +253,40 @@ static func add_custom_blocks(
 static func get_variable_block_definitions(variables: Array[VariableDefinition]) -> Array[BlockDefinition]:
 	var block_definitions: Array[BlockDefinition] = []
 	for variable: VariableDefinition in variables:
-		var type_string: String = Types.VARIANT_TYPE_TO_STRING[variable.var_type]
+		var block_def: BlockDefinition
 
 		# Getter
-		var block_def = BlockDefinition.new()
-		block_def.name = "get_var_%s" % variable.var_name
-		block_def.category = "Variables"
-		block_def.type = Types.BlockType.VALUE
-		block_def.variant_type = variable.var_type
-		block_def.display_template = variable.var_name
-		block_def.code_template = variable.var_name
+		block_def = get_variable_getter_block_definition(variable)
 		block_definitions.append(block_def)
 
 		# Setter
-		block_def = BlockDefinition.new()
-		block_def.name = "set_var_%s" % variable.var_name
-		block_def.category = "Variables"
-		block_def.type = Types.BlockType.STATEMENT
-		block_def.display_template = "Set %s to {value: %s}" % [variable.var_name, type_string]
-		block_def.code_template = "%s = {value}" % [variable.var_name]
+		block_def = get_variable_setter_block_definition(variable)
 		block_definitions.append(block_def)
 
 	return block_definitions
+
+
+static func get_variable_getter_block_definition(variable: VariableDefinition) -> BlockDefinition:
+	var block_def := BlockDefinition.new()
+
+	block_def.name = "get_var_%s" % variable.var_name
+	block_def.category = "Variables"
+	block_def.type = Types.BlockType.VALUE
+	block_def.variant_type = variable.var_type
+	block_def.display_template = variable.var_name
+	block_def.code_template = variable.var_name
+
+	return block_def
+
+
+static func get_variable_setter_block_definition(variable: VariableDefinition) -> BlockDefinition:
+	var type_string: String = Types.VARIANT_TYPE_TO_STRING[variable.var_type]
+	var block_def := BlockDefinition.new()
+
+	block_def.name = "set_var_%s" % variable.var_name
+	block_def.category = "Variables"
+	block_def.type = Types.BlockType.STATEMENT
+	block_def.display_template = "Set %s to {value: %s}" % [variable.var_name, type_string]
+	block_def.code_template = "%s = {value}" % [variable.var_name]
+
+	return block_def
--- a/addons/block_code/code_generation/blocks_catalog.gd
+++ b/addons/block_code/code_generation/blocks_catalog.gd
@@ -287,6 +287,18 @@ static func get_variable_setter_block_definition(variable: VariableDefinition) -
 	block_def.category = "Variables"
 	block_def.type = Types.BlockType.STATEMENT
 	block_def.display_template = "Set %s to {value: %s}" % [variable.var_name, type_string]
-	block_def.code_template = "%s = {value}" % [variable.var_name]
+	block_def.code_template = "%s = {value}" % variable.var_name
 
+	return block_def
+
+
+static func get_property_getter_block_definition(variable: VariableDefinition) -> BlockDefinition:
+	var block_def := get_variable_getter_block_definition(variable)
+	block_def.description = "The %s property" % variable.var_name
+	return block_def
+
+
+static func get_property_setter_block_definition(variable: VariableDefinition) -> BlockDefinition:
+	var block_def := get_variable_setter_block_definition(variable)
+	block_def.description = "Set the %s property" % variable.var_name
 	return block_def
--- a/addons/block_code/serialization/block_script_serialization.gd
+++ b/addons/block_code/serialization/block_script_serialization.gd
@@ -83,22 +83,26 @@ func instantiate_block_by_name(block_name: String) -> Block:
 
 func get_block_definition(block_name: String) -> BlockDefinition:
 	var split := block_name.split(":", true, 1)
+	var block_definition: BlockDefinition
 
 	if len(split) > 1:
 		return _get_parameter_block_definition(split[0], split[1])
 
-	var block_definition = _get_base_block_definition(block_name)
+	block_definition = _get_base_block_definition(block_name)
+	if block_definition != null:
+		return block_definition
 
-	if block_definition == null:
-		# FIXME: This is a workaround for old-style output block references.
-		#        These were generated ahead of time using a block name that has
-		#        a "_" before the parameter name. Now, these parameter blocks
-		#        are generated on demand for any block name containing a ":".
-		#        Please remove this fallback when it is no longer necessary.
-		split = block_name.rsplit("_", true, 1)
-		return _get_parameter_block_definition(split[0], split[1])
+	block_definition = _get_obj_property_block_definition(block_name)
+	if block_definition != null:
+		return block_definition
 
-	return block_definition
+	# FIXME: This is a workaround for old-style output block references.
+	#        These were generated ahead of time using a block name that has
+	#        a "_" before the parameter name. Now, these parameter blocks
+	#        are generated on demand for any block name containing a ":".
+	#        Please remove this fallback when it is no longer necessary.
+	split = block_name.rsplit("_", true, 1)
+	return _get_parameter_block_definition(split[0], split[1])
 
 
 func _get_base_block_definition(block_name: String) -> BlockDefinition:
@@ -133,6 +137,49 @@ func _get_parameter_block_definition(block_name: String, parameter_name: String)
 	block_definition.scope = base_block_definition.code_template
 
 	return block_definition
+
+
+func _get_obj_property_block_definition(block_name: String) -> BlockDefinition:
+	var block_definition: BlockDefinition
+	var variable: VariableDefinition
+	var property_name: String
+	var is_getter = true
+
+	if block_name.begins_with("get_var_"):
+		property_name = block_name.get_slice("get_var_", 1)
+	elif block_name.begins_with("set_var_"):
+		property_name = block_name.get_slice("set_var_", 1)
+		is_getter = false
+	else:
+		return null
+
+	# Getter block needs the property's variant type information by visiting the
+	# block_code_node's parent node because the type is not saved as a key of
+	# the resource in the scene file
+	var property_info := _get_parent_node_property_info(property_name)
+	if not property_info.has("type"):
+		return null
+
+	if is_getter:
+		variable = VariableDefinition.new(property_name, property_info["type"])
+		block_definition = BlocksCatalog.get_property_getter_block_definition(variable)
+	else:
+		variable = VariableDefinition.new(property_name, property_info["type"])
+		block_definition = BlocksCatalog.get_property_setter_block_definition(variable)
+
+	return block_definition
+
+
+func _get_parent_node_property_info(property_name: String) -> Dictionary:
+	if not block_code_node:
+		return {}
+
+	var properties := block_code_node.get_parent().get_property_list()
+	for property in properties:
+		if property["name"] == property_name:
+			return property
+
+	return {}
 
 
 func _update_block_definitions():
--- a/addons/block_code/serialization/block_script_serialization.gd
+++ b/addons/block_code/serialization/block_script_serialization.gd
@@ -28,6 +28,8 @@ const SCENE_PER_TYPE = {
 @export var generated_script: String
 @export var version: int
 
+var block_code_node: BlockCode
+
 var _available_blocks: Array[BlockDefinition]
 var _categories: Array[BlockCategory]
 
--- a/addons/block_code/ui/block_canvas/block_canvas.gd
+++ b/addons/block_code/ui/block_canvas/block_canvas.gd
@@ -81,8 +81,10 @@ func _can_drop_data(at_position: Vector2, data: Variant) -> bool:
 	if typeof(data) != TYPE_DICTIONARY:
 		return false
 
-	# Allow dropping property block
+	# Allow dropping property block of the block code node's parent node
 	if data.get("type", "") == "obj_property":
+		if data["object"] != _context.parent_node:
+			return false
 		return true
 
 	var nodes: Array = data.get("nodes", [])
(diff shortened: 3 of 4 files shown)


addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -131,37 +131,16 @@ func _drop_obj_property(at_position: Vector2, data: Variant) -> void:
...
-				"Variables",
-				Types.BlockType.STATEMENT,
-				property_type,
-				"set %s to {value: %s}" % [property_name.capitalize().to_lower(), type_string],
-				"%s = {value}" % property_name,
-				{"value": property_value},
-			)
-		)
+		block_name = &"%s_set_%s" % [object_name, property_name]
+		block_definition = _context.block_script.get_obj_property_block_definition(block_name, property_value)
```
wjt, line 139: I really don't like how we construct a string and then pass it to a function which immediately splits it back up into the 3 pieces we had here.

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -79,24 +79,29 @@ func instantiate_block_by_name(block_name: String) -> Block:
...
-		#        These were generated ahead of time using a block name that has
-		#        a "_" before the parameter name. Now, these parameter blocks
-		#        are generated on demand for any block name containing a ":".
-		#        Please remove this fallback when it is no longer necessary.
-		split = block_name.rsplit("_", true, 1)
-		return _get_parameter_block_definition(split[0], split[1])
+	var property_value = arguments.get("value", null)
+	block_definition = get_obj_property_block_definition(block_name, property_value)
+	if block_definition != null:
+		return block_definition
```
dylanmccall, line 96: Have you tried building this feature around `_get_parameter_block_definition`? There's a bit of overlap here. In fact they're solving the same fundamental problem: it's a type of block where we generate a block definition when we need it, based on an identifier such as a property name.

As it exists, that function only generates the one kind of value block, which is an output parameter for another block. (It needs a base block definition so it knows what output parameters are defined). So it needs extra help, but it should be feasible to extend it similarly to what you have in `get_obj_property_block_definition`, so our parameter blocks would start looking like _MyType_get:property_.

I think there's a few ways to make that work. I think ideally we'd have `_get_base_block_definition` generate a plain _MyType_get_ and _MyType_set_ block for whatever type is at hand, and we'd have something in BlockDefinition that `_get_parameter_block_definition` calls which would be like "give me a parameterized version of this block definition with foo set to bar".

I realize I'm asking for a fair bit of extra machinery here so I understand if we just want to land a bug fix, but I did have this case in mind when I was adding parameter block definitions. Sorry I didn't really document it :)

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -133,6 +138,43 @@ func _get_parameter_block_definition(block_name: String, parameter_name: String)
...
+	else:
+		var type_string: String = Types.VARIANT_TYPE_TO_STRING[property_type]
+		block_definition.name = block_name
+		block_definition.target_node_class = object_name
+		block_definition.description = "Set the %s property" % property_name
+		block_definition.category = "Variables"
+		block_definition.type = Types.BlockType.STATEMENT
+		block_definition.variant_type = property_type
+		block_definition.display_template = "set %s to {value: %s}" % [property_name.capitalize().to_lower(), type_string]
+		block_definition.code_template = "%s = {value}" % property_name
```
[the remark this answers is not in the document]
starnight: Interesting idea!

I referred to get_variable_block_definitions() actually.  But, object property and variable blocks have different block names:
* <object name>_(get|set)_<property name>
* (get|set)_var_<variable name>

Hm!  Let me try / test to use the variable one for object property.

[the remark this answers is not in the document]
starnight: Okay!  It can be.  Updated as new commits.

addons/block_code/serialization/block_script_serialization.gd
starnight: Let me study this more.

starnight: I try some scenario as examples: the "speed" and "floor_stop_on_slope" properties
* "Object_get:speed" will be split for `_get_parameter_block_definition()`
  - block_name: "Object_get"
  - parameter_name: "speed"
*  "Object_get:floor_stop_on_slope" will be split for `_get_parameter_block_definition()`
  - block_name: "Object_get"
  - parameter_name: "floor_stop_on_slope"

Because of the different `parameter_name`s, we have to generate different base block definitions for [`parent_out_parameters.has(parameter_name)`](https://github.com/endlessm/godot-block-coding/blob/c1193269ec0d655b97486419b45a10d78fa4c4f0/addons/block_code/serialization/block_script_serialization.gd#L117), which means the object property block's logic is different from the parameter block.

starnight: Besides, it loses the property's variant type after close the project actually.  Because, the property's variant type is not saved in the resource of `.tscn`.  The setter block's variant type can be restored by the stored **value**.  However, getter block does not have the value information.

starnight: Based on the limitation of the **parameter_name** and the getter's variant type, I think we cannot generate blocks via `_get_parameter_block_definition()` now.  So, I go back to the `_get_obj_property_block_definition()` in the updated commits.

starnight:
I encode the property's variant type into block name finally.  So, after reopen the project, block coding can restore the variant type information of the property's getter block.  Here is the demo video:
https://github.com/user-attachments/assets/82e9d524-0d5a-4a1d-947d-7af17a51314a

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -133,6 +138,33 @@ func _get_parameter_block_definition(block_name: String, parameter_name: String)
 	return block_definition
 
 
+func _get_obj_property_block_definition(block_name: String, property_value: Variant) -> BlockDefinition:
+	var block_definition: BlockDefinition
+	var variable: VariableDefinition
+	var property_name: String
+	var is_getter = true
+
+	if block_name.find("get_var_") == 0:
```
manuq, line 147:
```suggestion
	if block_name.begins_with("get_var_"):
```

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -133,6 +138,33 @@ func _get_parameter_block_definition(block_name: String, parameter_name: String)
...
+		property_name = block_name.get_slice("get_var_", 1)
+	elif block_name.find("set_var_") == 0:
```
manuq, line 149:
```suggestion
	elif block_name.begins_with("set_var_"):
```

manuq:
Trying to get the property name from the block name like `foo` from `get_var_foo` is not ideal. The VariableDefinition in `addons/block_code/code_generation/variable_definition.gd` already has name and type.

Now as discussed, reusing the `variables` in `BlockScriptSerialization` is a problem because properties shouldn't be declared as `var` in the script. And checking `ClassDB.class_get_property_list()` doesn't work for custom node's properties like the Simple ones. Two alternatives occur to me to separate variables from properties:
- Enhace VariableDefinition with an `is_property` boolean field
- Add `@export var properties: Array[VariableDefinition]` to the `BlockScriptSerialization` resource

addons/block_code/serialization/block_script_serialization.gd
dbnicholson: Sorry for coming in late here, but the property doesn't need to be encoded in the scene file because it's part of the object it's being applied to. It's more work, but you can call [get_property_list](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-method-get-property-list) on the target object to determine the type. I think that would be more elegant than encoding the type in the block name. It wouldn't be that hard to get the property list once at the beginning of deserializing and then find the appropriate property as the definitions are loaded.

dbnicholson:
> Now as discussed, reusing the `variables` in `BlockScriptSerialization` is a problem because properties shouldn't be declared as `var` in the script. And checking `ClassDB.class_get_property_list()` doesn't work for custom node's properties like the Simple ones. Two alternatives occur to me to separate variables from properties:
> 
> * Enhace VariableDefinition with an `is_property` boolean field
> * Add `@export var properties: Array[VariableDefinition]` to the `BlockScriptSerialization` resource

I don't see how adding an `is_property` field helps here since you still need to figure out what type of property it is. I'm also not crazy about exporting the property list since it's bound to get out of sync with the target node's properties.

I came up with a 3rd option. What we really want is to look at the target node's instance, but that's not available in the serialization since it's a child of the node. However, we can find the node from the scene tree because we know that the parent `BlockCode` node's `block_script` property will be set to this instance. In editor-only code without safety:

```
var root := EditorInterface.get_edited_scene_root()
for node in root.find_children("*", "BlockCode"):
    if node.block_script == self:
        var target_properties := node.get_parent().get_property_list()
```

https://github.com/endlessm/godot-block-coding/tree/block-script-target adds some helpers to do it. When deserializing property block definitions, you'd do something like:

```
var property := _get_target_property("thename")
var variant_type: Variant.Type = property["type"]
```

dbnicholson:
Another simpler possibility is that we have the `BlockCode` just tell `BlockScriptSerialization` about itself when it's instantiated:

```
diff --git a/addons/block_code/block_code_node/block_code.gd b/addons/block_code/block_code_node/block_code.gd
index 659719c..d08e89a 100644
--- a/addons/block_code/block_code_node/block_code.gd
+++ b/addons/block_code/block_code_node/block_code.gd
@@ -5,7 +5,15 @@ extends Node
 
 const TxUtils := preload("res://addons/block_code/translation/utils.gd")
 
-@export var block_script: BlockScriptSerialization = null
+@export var block_script: BlockScriptSerialization = null:
+       set(value):
+               # If the script is being cleared, remove ourself from the existing script.
+               if value == null:
+                       if block_script:
+                               block_script.block_code_node = null
+               else:
+                       value.block_code_node = self
+               block_script = value
 
 
 func _init():
diff --git a/addons/block_code/serialization/block_script_serialization.gd b/addons/block_code/serialization/block_script_serialization.gd
index 5f3051e..e0847f0 100644
--- a/addons/block_code/serialization/block_script_serialization.gd
+++ b/addons/block_code/serialization/block_script_serialization.gd
@@ -28,6 +28,8 @@ const SCENE_PER_TYPE = {
 @export var generated_script: String
 @export var version: int
 
+var block_code_node: BlockCode
+
 var _available_blocks: Array[BlockDefinition]
```

With the `BlockCode` node in hand, you can do all the same things like get the target object with `get_parent()` and then sift through its properties with `target.get_property_list()`.

starnight:
> I came up with a 3rd option. What we really want is to look at the target node's instance, but that's not available in the serialization since it's a child of the node. However, we can find the node from the scene tree because we know that the parent `BlockCode` node's `block_script` property will be set to this instance. In editor-only code without safety:
> 
> ```
> var root := EditorInterface.get_edited_scene_root()
> for node in root.find_children("*", "BlockCode"):
>     if node.block_script == self:
>         var target_properties := node.get_parent().get_property_list()
> ```

I was stuck at why I cannot `get_parent()`, which should be the BlockCode node's parent directly in the `addons/block_code/serialization/block_script_serialization.gd` until I learned [`block_script_serialization.gd` is loaded as a property ](https://github.com/endlessm/godot-block-coding/blob/c1193269ec0d655b97486419b45a10d78fa4c4f0/addons/block_code/block_code_node/block_code.gd#L32-L37)of the BlockCode node.  Thank you @dbnicholson

dbnicholson:
> I was stuck at why I cannot `get_parent()`, which should be the BlockCode node's parent directly in the `addons/block_code/serialization/block_script_serialization.gd` until I learned [`block_script_serialization.gd` is loaded as a property ](https://github.com/endlessm/godot-block-coding/blob/c1193269ec0d655b97486419b45a10d78fa4c4f0/addons/block_code/block_code_node/block_code.gd#L32-L37)of the BlockCode node. Thank you @dbnicholson

Just to clarify a bit. Only `Node`s like `BlockCode` are part of the scene tree and can navigate to their ancestors or descendents using things like `get_parent()`. `BlockScriptSerialization` is a `Resource`, which is not a `Node`. Godot doesn't provide any sort of methods like "what other objects hold a reference to this object".

addons/block_code/serialization/block_script_serialization.gd
starnight: Thanks!  That is more readable.  Fixed!

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -135,6 +140,47 @@ func _get_parameter_block_definition(block_name: String, parameter_name: String)
...
+		# the block_code_node's parent node because the type is not saved as a
+		# key of the resource in the scene file
+		var property_info := _get_parent_node_property_info(property_name)
+		if not property_info.has("type"):
+			return null
+
+		variable = VariableDefinition.new(property_name, property_info["type"])
+		block_definition = BlocksCatalog.get_property_getter_block_definition(variable)
+	else:
+		variable = VariableDefinition.new(property_name, typeof(property_value))
```
dbnicholson, line 168: Why not get the parent node property for the setter, too? Then you wouldn't need to pass the value in at all.

addons/block_code/code_generation/blocks_catalog.gd
```diff
@@ -287,6 +287,18 @@ static func get_variable_setter_block_definition(variable: VariableDefinition) -
...
-	block_def.code_template = "%s = {value}" % [variable.var_name]
+	block_def.code_template = "%s = {value}" % variable.var_name
 
 	return block_def
+
+
+static func get_property_getter_block_definition(variable: VariableDefinition) -> BlockDefinition:
+	var block_def := get_variable_getter_block_definition(variable)
+	block_def.description = "The %s property" % variable.var_name
+	return block_def
```
dbnicholson, line 298: It's a bit unfortunate that these property blocks will be in the Variables section, but I can't think of a better place at the moment.

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -28,6 +28,8 @@ const SCENE_PER_TYPE = {
 @export var generated_script: String
 @export var version: int
 
+var block_code_node: BlockCode
```
dbnicholson, line 31: A potential future improvement here would be to cache the parent properties when set so that `_get_parent_node_property_info` didn't have to iterate the entire property list every time. I don't think that's worth it right now, though.

dbnicholson (changes requested):
I think this makes sense to me. The only thing I'd want to see addressed is the change to pass in the block arguments when serializing. That doesn't seem necessary if you're already using the property information for the property getter block.

addons/block_code/serialization/block_script_serialization.gd
starnight: Oh! yes.  Good point.

dbnicholson (approved):
That looks like it should work! 2 final nitpicks, but feel free to merge after that.

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -135,6 +140,47 @@ func _get_parameter_block_definition(block_name: String, parameter_name: String)
...
+		block_definition = BlocksCatalog.get_property_getter_block_definition(variable)
+	else:
+		variable = VariableDefinition.new(property_name, property_info["type"])
+		block_definition = BlocksCatalog.get_property_setter_block_definition(variable)
+
+	return block_definition
+
+
+func _get_parent_node_property_info(property_name: String) -> Dictionary:
+	var properties := block_code_node.get_parent().get_property_list()
```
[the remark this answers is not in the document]
starnight: Hmm ...  I was thinking about this, too.  In theory, `_get_parent_node_property_info()` is only invoked after the script is loaded and attached to a block code node.  But, I think some evils live in the node/resource's life cycle transition time.

addons/block_code/serialization/block_script_serialization.gd
```diff
@@ -81,24 +81,29 @@ func instantiate_block_by_name(block_name: String) -> Block:
 	return instantiate_block(block_definition)
 
 
-func get_block_definition(block_name: String) -> BlockDefinition:
+func get_block_definition(block_name: String, arguments: Dictionary) -> BlockDefinition:
```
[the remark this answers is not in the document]
starnight: oh, right.  The `arguments` is unused now.  Thanks!

addons/block_code/serialization/block_script_serialization.gd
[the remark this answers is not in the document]
starnight: Fixed in the new commit.

starnight:
Thank you guys' help!  I merge this!

PR #319, opened 2024-11-08
Update GitHub issue templates

manuq:
This is an open-source plugin and now also developed in the open. Remove link to Endless Employees issue tracker. Edit the "public" template to be the sole issue template.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,5 +1,0 @@
-blank_issues_enabled: false
-contact_links:
-  - name: Create Internal Task
-    url: https://phabricator.endlessm.com/tag/baby_godot/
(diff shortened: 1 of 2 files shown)


manuq:
Honestly I'm new to GitHub templates so I'm not sure if it will work. These were added by @cassidyjames in commit 6f7bb2d7 . Reading https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser it looks like by removing the c`config.yml` the option for Endless employees should be removed.

manuq:
The chooser is still there but now with the only option we want: https://github.com/endlessm/godot-block-coding/issues/new/choose

Maybe in the future we can have a "feature proposal" template? With a field to justify the educational purposes that the README describes.

PR #320, opened 2024-11-12
Helper block for collision with group

manuq:
This is a 3-in-1 helper block for the common pattern found in the demos. It replaces adding:
- A "when this node collides with ..." block
- An "if" block
- An "is node in group" block


manuq:
Draft for now because our signal handlers have the same "on_SIGNAL" name.

urbit-pilled:
Having two of these blocks will generate 2 functions with the same "_on_body_entered" callback, is this the problem?

If so, doesn't this problem already exist in the "when this node collides with .." block?

This will be a hard fix because you need a way to make the signal callback unique but the user codes the function name with `code_template` so in order for it to be unique it needs to be dynamically generated.

manuq:
> Having two of these blocks will generate 2 functions with the same "_on_body_entered" callback, is this the problem you're talking about with the "on_SIGNAL" name?

Yes, we need a better way. For example the Visual Shaders in Godot use suffixes (try creating a visual shader and then click "Show generated shader code". We'll need a solution like that.

Issue #279, opened 2024-10-18
Change shape of booleans to be hexagonal


manuq:
### What Happened

This is a nice convention of Scratch and other block-coding languages. Boolean blocks and the slots that expect booleans are hexagonal or "pointy". This helps the user to understand where to add booleans.

Proposal: change the shape of value blocks to be pointy when the variant type is boolean. Same for the shape of slots that expect a boolean.

Here is how they look today:
![Captura desde 2024-10-18 16-28-57](https://github.com/user-attachments/assets/bc43580c-62f7-455f-a67f-c2bd1273c8c1)

And this is a reference from Scratch: https://www.codingkids.com.au/wp-content/uploads/2017/07/img1.png

### How to Reproduce

N/A, this is a feature proposal.

### Block Coding Plugin Version

main

Issue #278, opened 2024-10-18
Change shape of entry blocks to make them unique


manuq:
### What Happened

Entry blocks are too similar to statement blocks, and this is confusing:
![Captura desde 2024-10-18 16-10-17](https://github.com/user-attachments/assets/b04b453b-2710-4833-a96b-4169224e9063)

Proposal: do like Scratch, add a half ellipse shape to entry blocks:
![when-starting](https://github.com/user-attachments/assets/1a1e91f0-bb02-4407-84b8-317fe573b520)

Here is a screenshot of Scratch for more reference: https://upload.wikimedia.org/wikipedia/commons/2/2c/Scratch3-moveasprite.png

### How to Reproduce

N/A, this is a feature proposal.

### Block Coding Plugin Version

main

PR #325, opened 2024-12-02
UI: Improve block shapes

manuq:
- Make entry blocks more distinguishable from the parameter blocks.
- Change boolean blocks to be pointy.

Contribution from @DoomTas3r https://github.com/endlessm/godot-block-coding/pull/273

Fix https://github.com/endlessm/godot-block-coding/issues/278
Fix https://github.com/endlessm/godot-block-coding/issues/279


addons/block_code/ui/blocks/parameter_block/parameter_block.gd
```diff
@@ -15,6 +15,18 @@ var spawned_by: ParameterOutput
 func _ready():
 	super()
 	_background.color = color
+	_update_block_shape()
+
+
+func _on_definition_changed():
+	super()
+	_update_block_shape()
```
starnight, line 23: The **_background** accessed in `_update_block_shape()` is only available when the node has been ready.  However, seems like `_on_definition_changed()` will be invoked by some where before this node becomes ready and shows error:
```
res://addons/block_code/ui/blocks/parameter_block/parameter_block.gd:29 - Invalid assignment of property or key 'is_pointy_value' with value of type 'bool' on a base object of type 'Nil'.
```
I think that here should add a if condition checking [is_node_ready()](https://docs.godotengine.org/en/stable/classes/class_node.html#class-node-method-is-node-ready) before `_update_block_shape()`.

starnight (changes requested):
Thanks arrange this improvement.  It makes more like logic flowchart! 
However, I notice a bug and place suggestion.

addons/block_code/ui/blocks/parameter_block/parameter_block.gd
manuq: Good catch @starnight ! I appended a fixup commit. Adding `await ready` did the trick.

PR #328, opened 2024-12-05
Pong example: Use blocks for sound resources

manuq:
Done by using the new feature, dragging the sound resources from the Filesystem dock to the block canvas.

The diff is crazy because resource IDs change all the time, but this is what it looks like now:
Captura desde 2024-12-05 12-14-34


wjt:
I ignored all the resource junk and looked at the generated code diff. Looks about right!

PR #331, opened 2024-12-06
Background: Disable/enable conditional properties

manuq:
The "is_pointy_value" property depends on the block type to be Value. And the "control_part" property depends on the block type to be Control.

[Grabación de pantalla desde 2024-12-06 15-24-47.webm](https://github.com/user-attachments/assets/3af886c0-c389-4c28-a8f0-d2a1b50ee7db)

### DIFF
--- a/addons/block_code/ui/blocks/utilities/background/background.gd
+++ b/addons/block_code/ui/blocks/utilities/background/background.gd
@@ -37,6 +37,7 @@ func _set_color(new_color):
 func _set_block_type(new_block_type):
 	block_type = new_block_type
 	queue_redraw()
+	notify_property_list_changed()
 
 
 func _set_control_part(new_control_part):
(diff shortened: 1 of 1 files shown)


manuq:
This is not important at all, but something I learned about today and wanted to put it somewhere.

addons/block_code/ui/blocks/utilities/background/background.gd
```diff
@@ -49,6 +50,13 @@ func _set_is_pointy_value(new_is_pointy_value):
 	queue_redraw()
 
 
+func _validate_property(property: Dictionary):
```
starnight, line 53: Simply place related link from Godot document [_validate_property()](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-private-method-validate-property)

addons/block_code/ui/blocks/utilities/background/background.gd
```diff
@@ -37,6 +37,7 @@ func _set_color(new_color):
 func _set_block_type(new_block_type):
 	block_type = new_block_type
 	queue_redraw()
+	notify_property_list_changed()
```
starnight, line 40: Simply place related link from Godot document [notify_property_list_changed()](https://docs.godotengine.org/en/stable/classes/class_object.html#class-object-private-method-validate-property)

PR #336, opened 2024-12-16
Added collapse/expand blocks for "advanced" settings

### LINKED ISSUE #288: Collapse/expand blocks for "advanced" settings ($200 bounty)
### What Happened

As we improve the blocks, more settings are being appended to them. These are usually left with the defaults and intended for advanced usage. For example:
Captura desde 2024-10-23 13-40-44

Proposal: allow blocks to expand/collapse to hide the advanced settings. Blocks should be collapsed by default.

Here is how Microsoft MakeCode Arcade does it:
Captura desde 2024-10-23 13-32-43
Captura desde 2024-10-23 13-32-57

Implementation: there should be a (+) button at the end of collapsed blocks, and a (-) button at the end of expanded blocks. The format of the block label template should be enhaced to define where the cut should happen. Maybe with a `|` character? So the label template of the sound block above would look like:
```
play the sound {name: STRING} | with volume {db: FLOAT} dB and pitch scale {pitch: FLOAT}
```
Pay attention to the `|` that is separating how it should look collapsed `play the sound {name: STRING}` and the advanced stuff (volume and pitch scale settings).

### How to Reproduce

N/A this is a feature proposal.

### Block Coding Plugin Version

_No response_

### Godot Engine Version

_No response_

### Operating System

_No response_

### Logs or command-line output

_No response_

### Anything else?

_No response_

joaquin30:
/claim #288
Closes #288. This adds a button for expand/collapse settings in blocks that are after the "|" separator.


manuq:
@joaquin30 looking good so far!

[Grabación de pantalla desde 2024-12-16 09-28-57.webm](https://github.com/user-attachments/assets/a47cc33f-5100-467d-8679-123656caab4b)

I know this is still draft, but for the 1st review make sure that this pass the linting and formatting check above. You can do that locally if you setup [pre-commit as in the README](https://github.com/endlessm/godot-block-coding?tab=readme-ov-file#pre-commit)

addons/block_code/ui/blocks/utilities/template_editor/collapsable_settings.gd
```diff
@@ -0,0 +1,36 @@
...
+extends HBoxContainer
+
+var collapsed = false
+
+
+# Called when the node enters the scene tree for the first time.
+func _ready() -> void:
+	collapse()
+	$CollapseButton.connect("button_up", collapse)
+	$ExpandButton.connect("button_up", expand)
```
manuq, line 12: For these buttons, please declare the nodes at the top of the script. And also add unique names to them, so the script doesn't depend on UI changes:
```
@onready var _collapse_button: Button = %CollapseButton
@onready var _expand_button: Button = %ExpandButton
```

addons/block_code/ui/blocks/utilities/template_editor/template_editor.gd
```diff
@@ -87,26 +88,36 @@ func _update_from_format_string():
 		child.queue_free()
 
 	var match_id = 0
+	var is_collapsable = false
+	var collapsable := CollapsableSettingsScene.instantiate() as CollapsableSettings
 	for item in BlockDefinition.parse_display_template(format_string):
+		var hidden = item.has("hidden") and item.get("hidden")
```
manuq, line 94: `item.get("hidden")` already returns `null` if the key is missing, and you can also pass a default. So you can probably do the same without the `has()` check:
```suggestion
		var hidden = item.get("hidden", false)
```
Untested, please confirm!

addons/block_code/ui/blocks/utilities/template_editor/template_editor.gd
```diff
@@ -87,26 +88,36 @@ func _update_from_format_string():
...
+			_append_label(collapsable if hidden else _container, item.get("label"))
 		elif item.has("in_parameter"):
-			_append_input_parameter(item.get("in_parameter"), match_id)
+			_append_input_parameter(collapsable if hidden else _container, item.get("in_parameter"), match_id)
 		elif item.has("out_parameter"):
-			_append_output_parameter(item.get("out_parameter"), match_id)
+			_append_output_parameter(collapsable if hidden else _container, item.get("out_parameter"), match_id)
 		elif item.has("const_parameter"):
-			_append_const_parameter(item.get("const_parameter"), match_id)
+			_append_const_parameter(collapsable if hidden else _container, item.get("const_parameter"), match_id)
```
manuq, line 104: The `collapsable if hidden else _container` is repeated 4 times. Is better to move it to a variable before the if/elif/elif/elif block.

Issue #257, opened 2024-10-10
Move with Player 1/2 block does not work with gamepads ($200 bounty)


cassidyjames:
### What Happened

I believe we currently hardcode the WASD and Up/Left/Down/Right buttons for the input. It'd be nice to be able to use gamepads with Block Coding, and Justin has said the Learning team would love that for the learning programs.

### How to Reproduce

1. Use the Move With Controls block
2. Plug in a controller
3. Be sad :(

PR #337, opened 2024-12-16
Added advanced mode

urbit-pilled:
/claim #282 

- adds a checkbox to the main_panel title bar
- adds `is_advanced` boolean to the block_definition (default value is false)
- hides advanced blocks when the checkbox is not ticked.

PR #338, opened 2024-12-16
Added controller support for move block

urbit-pilled:
/claim #257 

I added some Input Actions to the InputMap so that `player_1_up` contains both keyboard and controller events.
Player 1 and Player 2 joystick buttons are identical aside from belonging to different devices. (Player 1 = Device 0 and Player 2 = Device 1)

### DIFF
--- a/addons/block_code/simple_nodes/simple_character/simple_character.gd
+++ b/addons/block_code/simple_nodes/simple_character/simple_character.gd
@@ -29,6 +29,36 @@ const PLAYER_KEYS = [
 	}
 ]
 
+const PLAYER_JOYSTICK_BUTTONS = {
+	"up": JOY_BUTTON_DPAD_UP,
+	"down": JOY_BUTTON_DPAD_DOWN,
+	"left": JOY_BUTTON_DPAD_LEFT,
+	"right": JOY_BUTTON_DPAD_RIGHT,
+}
+
+const PLAYER_JOYSTICK_MOTION = {
+	"up":
+	{
+		"axis": JOY_AXIS_LEFT_Y,
+		"axis_value": -1,
+	},
+	"down":
+	{
+		"axis": JOY_AXIS_LEFT_Y,
+		"axis_value": 1,
+	},
+	"left":
+	{
+		"axis": JOY_AXIS_LEFT_X,
+		"axis_value": -1,
+	},
+	"right":
+	{
+		"axis": JOY_AXIS_LEFT_X,
+		"axis_value": 1,
+	},
+}
+
 var sprite: Sprite2D
 var collision: CollisionShape2D
 
@@ -99,6 +129,19 @@ func _setup_actions():
 			var e = InputEventKey.new()
 			e.physical_keycode = PLAYER_KEYS[i][action]
 			InputMap.action_add_event(action_name, e)
+
+			#controller d-pad event
+			var ej = InputEventJoypadButton.new()
+			ej.device = i
+			ej.button_index = PLAYER_JOYSTICK_BUTTONS[action]
+			InputMap.action_add_event(action_name, ej)
+
+			#controller left stick event
+			var ejm = InputEventJoypadMotion.new()
+			ejm.device = i
+			ejm.axis = PLAYER_JOYSTICK_MOTION[action]["axis"]
+			ejm.axis_value = PLAYER_JOYSTICK_MOTION[action]["axis_value"]
+			InputMap.action_add_event(action_name, ejm)
 
 
 func get_custom_class():
--- a/addons/block_code/simple_nodes/simple_character/simple_character.gd
+++ b/addons/block_code/simple_nodes/simple_character/simple_character.gd
@@ -14,22 +14,20 @@ const Types = preload("res://addons/block_code/types/types.gd")
 
 @export var speed: Vector2 = Vector2(300, 300)
 
-const PLAYER_KEYS = {
-	"player_1":
+const PLAYER_KEYS = [
 	{
 		"up": KEY_W,
 		"down": KEY_S,
 		"left": KEY_A,
 		"right": KEY_D,
 	},
-	"player_2":
 	{
 		"up": KEY_UP,
 		"down": KEY_DOWN,
 		"left": KEY_LEFT,
 		"right": KEY_RIGHT,
 	}
-}
+]
 
 var sprite: Sprite2D
 var collision: CollisionShape2D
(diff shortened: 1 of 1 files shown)


addons/block_code/simple_nodes/simple_character/simple_character.gd
```diff
@@ -31,6 +31,27 @@ const PLAYER_KEYS = {
...
+	"player_2":
+	{
+		1: {
+			"up": JOY_BUTTON_DPAD_UP,
+			"down": JOY_BUTTON_DPAD_DOWN,
+			"left": JOY_BUTTON_DPAD_LEFT,
+			"right": JOY_BUTTON_DPAD_RIGHT,
+		}
+	}
+}
```
wjt, line 53: I understand that you've done this for consistency with the keyboard actions above but I don't think you need two copies of the actions: I think you could do:

```
const JOYPAD_BUTTONS = {
    "up": JOY_BUTTON_DPAD_UP,
   ...
}
```

and iterate it once per player.

addons/block_code/simple_nodes/simple_character/simple_character.gd
```diff
@@ -83,24 +104,44 @@ func _ready():
...
+	InputMap.add_action("player_1_down")
+	InputMap.add_action("player_2_left")
+	InputMap.add_action("player_2_right")
+	InputMap.add_action("player_2_up")
+	InputMap.add_action("player_2_down")
+
+	for player in PLAYER_KEYS:
+		for action in PLAYER_KEYS[player]:
+			var e = InputEventKey.new()
+			e.keycode = PLAYER_KEYS[player][action]
```
wjt, line 127: Previously the input vector was calculated with `Input.is_physical_key_pressed` which corresponds to `InputEventKey.physical_keycode`. See commit 16ab468742f47901174e43ab19608380ba6d8c52 for why this is significant. Please adjust this to use `physical_keycode` instead.

addons/block_code/simple_nodes/simple_character/simple_character.gd
```diff
@@ -83,24 +104,44 @@ func _ready():
...
 	add_to_group("affected_by_gravity", true)
+	_setup_actions()
 	_texture_updated()
 
 
-func get_custom_class():
-	return "SimpleCharacter"
+func _setup_actions():
+	if(InputMap.has_action("player_1_left")):
+		return
```
wjt, line 113: I was concerned that this would run in the editor (not just when the game is launched) and edit the project's input map. That would be problematic because it would mean that subsequent changes here would not be applied to the game. But in fact this works as desired: it is not run in the editor. (I don't actually know why because this is a `@tool`...)

addons/block_code/simple_nodes/simple_character/simple_character.gd
```diff
@@ -31,6 +31,27 @@ const PLAYER_KEYS = {
 	}
 }
 
+const PLAYER_JOYSTICK_BUTTONS = {
+	"player_1": {
+		0: {
+			"up": JOY_BUTTON_DPAD_UP,
```
wjt, line 37: Please also map the left stick of each controller to these actions. From my experience play-testing with younger children, they expect the joystick to work and don't use the d-pad.

wjt (changes requested):
On top of the changes requested below, as mentioned in https://github.com/endlessm/godot-block-coding/pull/337 please fix the linter errors. You can set up and run `pre-commit` locally as documented in the README.

(For what it's worth, I really don't like how `gdformat` formats nested dictionaries but that's the price of standardization.)

addons/block_code/simple_nodes/simple_character/simple_character.gd
urbit-pilled: right, that would be much cleaner. fixed in c502b4a

addons/block_code/simple_nodes/simple_character/simple_character.gd
urbit-pilled: good catch, I fixed it as part of 8be0a70

addons/block_code/simple_nodes/simple_character/simple_character.gd
urbit-pilled: added left joystick controls in 8be0a70

urbit-pilled:
thanks for the review @wjt, I just fixed all the changes so it should be ready for re-review now.

addons/block_code/simple_nodes/simple_character/simple_character.gd
```diff
@@ -31,6 +31,36 @@ const PLAYER_KEYS = {
...
+	"up":
+	{
+		"axis": JOY_AXIS_LEFT_Y,
+		"axis_value": 1,
+	},
+	"down":
+	{
+		"axis": JOY_AXIS_LEFT_Y,
+		"axis_value": -1,
+	},
```
wjt, line 51: These are upside-down: pushing the joystick up moves the character down.

addons/block_code/simple_nodes/simple_character/simple_character.gd
```diff
@@ -83,24 +113,46 @@ func _ready():
...
+			InputMap.action_add_event(action_name, ejm)
+
+
+func get_custom_class():
+	return "SimpleCharacter"
 
 
 func move_with_player_buttons(player: String, kind: String, delta: float):
-	var direction = _player_input_to_direction(player)
+	var direction = Input.get_vector(player + "_left", player + "_right", player + "_up", player + "_down")
```
wjt, line 155: In the `"platformer"` block below, `PLAYER_KEYS` is still hardcoded – you need to adjust this to check the up action instead. Please forgive me missing this in previous rounds of review.

addons/block_code/simple_nodes/simple_character/simple_character.gd
```diff
@@ -83,24 +113,46 @@ func _ready():
...
+func _setup_actions():
+	if Engine.is_editor_hint() or InputMap.has_action("player_1_left"):
+		return
+
+	for player in PLAYER_KEYS:
+		for action in PLAYER_KEYS[player]:
+			var action_name = player + "_" + action
+			InputMap.add_action(action_name)
+
+			var device = int(player.split("_")[-1]) - 1
```
wjt, line 129: This is OK, but it might be clearer to change `PLAYER_KEYS` to an array of dictionaries, then iterate it with something like this:

```gdscript
for i in PLAYER_KEYS.size():
    for action in PLAYER_KEYS[i]:
        var action_name = "player_%d_%s" % [i + 1, action]
        # etc.
```

You're still having to do arithmetic to adjust between 0-based array/controller indices and 1-based human-readable indices, but you wouldn't have to do this string mangling & conversion back to int.

addons/block_code/simple_nodes/simple_character/simple_character.gd
wjt: Annoyingly there isn't a platformer example in this repo. There is one in https://github.com/endlessm/godot-block-coding-demos, but you can also test it in the pong example by adjusting the block code as follows:

image

(It is the "set speed to Vector2(0, 1000)" which prevents the paddles from moving left and righ.t)

wjt (changes requested):

addons/block_code/simple_nodes/simple_character/simple_character.gd
urbit-pilled: fixed in commit dafe38e

addons/block_code/simple_nodes/simple_character/simple_character.gd
urbit-pilled: Thanks for the tip @wjt.

I just tested the platformer code and it should work now. I replaced the hardcoded `PLAYER_KEYS` with `direction.y`

Relevant commit: fd7db2b

urbit-pilled:
Thanks for the review @wjt, I just made the relevant changes, so it should be ready for a re-review now.

wjt (approved):
Looks great and works in my testing. Thanks for this contribution! I took the liberty of squashing the branch down to two commits: one to rework how keyboard input is handled, and another to add controller input.

wjt:
@dsd please can you award the bounty to this contributor?

PR #339, opened 2024-12-16
/claim #257

Gyan-max:
/claim #257
 Problem Overview
The "Move with Player 1/2" block currently only works with keyboard inputs (e.g., WASD keys and arrow keys).
The goal is to extend this functionality so that the block can also work with gamepad inputs (such as analog stick movements or D-pad buttons) to allow for more flexible control options, especially for users who prefer using a gamepad.
2. Solution Overview
The solution likely involves modifying the existing block logic to incorporate gamepad support. Here's how that could have been done:

Input Type Handling: A new input_type parameter or method is added, which allows the system to determine whether the inputs should come from the keyboard or gamepad.
Gamepad Input Mapping: The standard Godot input system can be used to map gamepad buttons and analog stick movements. In Godot, the Input class can be used to detect gamepad inputs, such as:
D-pad or Analog Stick: Detect movement from the analog sticks or D-pad.
Gamepad Buttons: Detect button presses (e.g., A, B, X, Y on controllers).
Controller Input Logic: The code in the move_with_player_buttons method would be updated to check for gamepad inputs in addition to keyboard inputs, possibly by using Input.get_joy_axis() for joystick movement or Input.is_joy_button_pressed() for button presses.

### DIFF
--- a/addons/block_code/simple_nodes/simple_character/simple_character.gd
+++ b/addons/block_code/simple_nodes/simple_character/simple_character.gd
@@ -101,15 +101,15 @@ func _player_input_to_direction(player: String):
 
 	# Gamepad input (new)
 	if player == "player_1" or player == "player_2":
-		var joy_index = 0 if player == "player_1" else 1 # Assuming player 1 uses the first gamepad and player 2 uses the second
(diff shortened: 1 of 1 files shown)


wjt:
Thank you for this submission but I have just accepted a preexisting implementation from #338.

PR #342, opened 2024-12-17
Adds the every physics step block

manuq:
And set it as an advanced block.

Fixes https://github.com/endlessm/godot-block-coding/issues/261


manuq:
Note: this is #261 just adding the new property "is_advanced" to true so it is filtered in the picker by default.

PR #346, opened 2024-12-18
Set Godot 4.3 as minimum supported version

wjt:
Godot 4.3 has been available for 4 months and it seems unlikely that anyone will have a pressing need to add blocks to a Godot 4.2 project.

Although we have been running unit tests against 4.2 in CI, those do not cover the full functionality of the plugin. To the best of my knowledge, no-one is regularly testing the plugin interactively against Godot 4.2.

Set the minimum version in the asset library template. Drop 4.2 from the CI test matrix. Remove some workarounds for 4.2 compatibility.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -33,7 +33,6 @@ jobs:
     strategy:
       matrix:
         godot-version:
-          - 4.2.2
           - 4.3.0
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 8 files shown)


manuq (changes requested):
Could you also address the "Polyfill of Node.is_part_of_edited_scene(), available to GDScript in Godot 4.3+." in `addons/block_code/ui/util.gd` ?

wjt:
Done

manuq (approved):

PR #347, opened 2024-12-18
SimpleCharacter: Fix axis confusion in tooltip

wjt:
Previously this description claimed that x is vertical and y horizontal.

Fix this, and clarify some other terms.


manuq (approved):
Good catch! And the other improvements are very welcome too.

PR #348, opened 2024-12-21
Condenses as vector2_xy and vector3_xyz blocks

DoomTas3r:
Condenses the existing vector2_x/y blocks and vector3_x/y/z blocks into singular blocks.

This makes switching an axis easier and reduces redundancy in the block picker.

Breaks compatibility with version 0.8.0

image


manuq (approved):
This is fantastic, you read my mind @DoomTas3r! I wanted to do this myself.

Issue #351, opened 2025-01-02
Integer to string casting and more string blocks


DoomTas3r:
Code_It_All made [this video](https://www.youtube.com/watch?v=FbZnldJe0XU) using version 0.8.0, but could not cast an integer to a string.

Perhaps make all string parameters accept only strings or improve implicit casting
`Parser Error: Value of type "int" cannot be assigned to a variable of type "String".`
![image](https://github.com/user-attachments/assets/c759af8c-50b8-4c43-844f-01cae3c29677)

A Text category could be added for these string-related blocks:

1. `(float) to string`
2. `length of (string)`
3. `letter (integer) of (string)`

And maybe more complex blocks:

5. `replace (string1) with (string2) in (string3)`
6. `slice (integer) of (string1) split by (string2)`

And finally, the concat block's label could change to differentiate from the add block, for example: `join (string1) (string2)`

The current concat block allows both strings and integers to be added, but it compiles to `string + integer`, which produces an error
`Parser Error: Invalid operands "String" and "int" for "+" operator.`
![image](https://github.com/user-attachments/assets/aa303c69-cfe9-49ae-be92-d0caab3cd348)

manuq:
> Code_It_All made [this video](https://www.youtube.com/watch?v=FbZnldJe0XU) using version 0.8.0, but could not cast an integer to a string.

First off, let's fix the bug. PR https://github.com/endlessm/godot-block-coding/pull/352 addresses the issue seen in the video.

On top of that, more text blocks would be welcome. Here is my attempt at "to string" which works with integers and floats. Ideally we would allow any type in its slot. I guess we could make the parameter type optional in `display_template` to allow any type: `{value} to string` instead of `{value: INT} to string`.

```
[gd_resource type="Resource" load_steps=2 format=3 uid="uid://d2eq3sdk61wkk"]

[ext_resource type="Script" path="res://addons/block_code/code_generation/block_definition.gd" id="1_g3t5i"]

[resource]
script = ExtResource("1_g3t5i")
name = &"to_string"
target_node_class = ""
description = "Format the value to String.

Advanced: A format string can be specified."
category = "Log"
type = 3
variant_type = 4
display_template = "{value: INT} to string | with format {format: STRING}"
code_template = "{format} % {value}
"
defaults = {
"format": "\"%s\""
}
signal_name = ""
scope = ""
is_advanced = true
```

DoomTas3r:
[...]

PR #352, opened 2025-01-06
Fix types relationships

manuq:
The types relationship (previously named cast_relationships) tell which type can be used in the context of another type. For example, integers and floats are exchangeable. And in particular, all other types can be used in boolean context: `if not Color.BLACK:` is allowed in GDScript as Color.BLACK is falsey. Same for the empty string, zero number, etc.

This fixes a bug where an integer or float value block was allowed to be snapped into a string slot.

This removes outdated custom casting code that has been abandoned.

https://github.com/endlessm/godot-block-coding/issues/351


manuq:
@DoomTas3r could you review? For some reason I can't add you to the reviewers list.

addons/block_code/types/types.gd
```diff
@@ -36,147 +36,14 @@ const STRING_TO_VARIANT_TYPE: Dictionary = {
...
-			continue
-
-		for edge in cast_graph[u]:
-			var v = edge.to
-			var alt = dist[u] + 1
-			if alt < dist[v]:
-				dist[v] = alt
-				prev[v] = CastGraphEdge.new(u, edge.cast_format)
-				queue.update_priority(v, alt)
+const types_relationships = {
```
DoomTas3r, line 39: I think this looks much cleaner, although I found casting to a boolean works in if statements or converting through a boolean block, but not for directly setting a boolean variable (`Parser Error: Value of type "String" cannot be assigned to a variable of type "bool".`) but that situation is unlikely due to it's unique shape. I also think that boolean to integer could be added in the future, similar to https://github.com/endlessm/godot-block-coding/issues/267 (true => 1, false => 0).

DoomTas3r (approved):
Works as intended by preventing casting as strings from other types.

addons/block_code/types/types.gd
manuq: > I think this looks much cleaner, although I found casting to a boolean works in if statements or converting through a boolean block, but not for directly setting a boolean variable (`Parser Error: Value of type "String" cannot be assigned to a variable of type "bool".`) but that situation is unlikely due to it's unique shape. I also think that boolean to integer could be added in the future, similar to #267 (true => 1, false => 0).

Oh right, good catch. The assignment is a problem. The issue was there before. For example:
Captura desde 2025-01-06 11-59-05

Which gives: `Parser Error: Value of type "Vector2" cannot be assigned to a variable of type "bool".`

This PR makes it a bit worse, allowing for more value blocks to be snapped bool setters.

manuq:
Marking this as draft because of https://github.com/endlessm/godot-block-coding/pull/352#discussion_r1904260061

Issue #357, opened 2025-01-23
Block Picker: Text input should not be editable


tindrew:
### What Happened

There's an issue with blocks. If you type a number into one of the blocks on the left side, the number persists even through the block being deleted. Example: 

![Image](https://github.com/user-attachments/assets/cc82b9c8-28d2-4eb8-b18a-2c008031e693)

After deletion it's back:

![Image](https://github.com/user-attachments/assets/136b8870-7121-4716-8946-78e40d19b37d)

You can see it on premade blocks too: 

![Image](https://github.com/user-attachments/assets/5d0f8762-7c1a-4257-854f-a04645de19a5)

It even persists if you add a BlockCode to a completely different node: 

![Image](https://github.com/user-attachments/assets/8345798a-956a-4b87-a7ed-c4ea2e6be339)

Also, if you drag that same block to the right side, the number disappears:

![Image](https://github.com/user-attachments/assets/1d94ac9d-ae1e-4106-926e-9f88677300e8)

### How to Reproduce

1. Add a couple nodes to a scene
2.  Add a block to one
3. Type a number into one of the existing blocks on the left hand side
4. delete the BlockCode completely.
5. Add a BlockCode to a different node.
6. Check the blocks, you'll see the number is still there.
7. Drag the block with the changed number to the right side. observe that the number disappears.

### Block Coding Plugin Version

v0.9.0

### Godot Engine Version

v4.3.0 Stable

### Operating System

Windows 10

### Logs or command-line output

```shell

```

PR #369, opened 2025-01-29
Make blocks in Picker not editable

manuq:
Add a new "editable" boolean property to Block, TemplateEditor and ParameterInput. When the block is in the picker, it will not be editable. Only when they are dragged to the canvas they become editable.

The property has to be in each of these components because Block can't access its ParameterInputs directly, only through its TemplateEditor.

The dropdown options in a not-editable block can still be changed to avoid disabling the button. But the change doesn't trigger any effect.

The default is editable = true, so blocks in the canvas are editable.

Fix https://github.com/endlessm/godot-block-coding/issues/357

### DIFF
--- a/addons/block_code/ui/blocks/block/block.gd
+++ b/addons/block_code/ui/blocks/block/block.gd
@@ -10,6 +10,10 @@ signal modified
 ## Color of block (optionally used to draw block color)
 @export var color: Color = Color(1., 1., 1.)
 
+## Whether the parameter inputs inside the block can be edited.
+@export var editable: bool = true:
+	set = _set_editable
+
 # FIXME Note: This used to be a NodePath. There is a bug in Godot 4.2 that causes the
 # reference to not be set properly when the node is duplicated. Since we don't
 # use the Node duplicate function anymore, this is okay.
@@ -54,6 +58,11 @@ func _ready():
 	focus_entered.connect(_block_on_focus_entered)
 	focus_exited.connect(_block_on_focus_exited)
 	_on_definition_changed()
+
+
+func _set_editable(value) -> void:
+	editable = value
+	template_editor.editable = value
 
 
 func _block_on_focus_entered():
--- a/addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
+++ b/addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
@@ -56,6 +56,22 @@ var _drag_start: Vector2 = Vector2.INF
 	_v3_z_line_edit: "",
 }
 
+## Whether the text inputs can be edited, and whether the options selection have any effect.
+@export var editable: bool = true:
+	set = _set_editable
+
+
+func _set_editable(value) -> void:
+	if not is_node_ready():
+		await ready
+	editable = value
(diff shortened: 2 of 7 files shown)


addons/block_code/ui/blocks/block/block.gd
```diff
@@ -102,6 +111,7 @@ func _update_template_editor():
 
 	template_editor.format_string = _get_format_string()
 	template_editor.parameter_defaults = _get_parameter_defaults()
+	# template_editor.editable = editable
```
wjt: Did you mean to leave this commented-out line here?

addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
```diff
@@ -180,6 +196,8 @@ func get_snapped_block() -> Block:
 
 
 func _validate_and_submit_edit_text(line_edit: Node, type: Variant.Type):
+	if not editable:
+		return
```
wjt, line 200: It seems weird to me that this is necessary when `_line_edit.editable` is set to match `editable` in `_set_editable`. I can see that if I focus (e.g.) the text field in the "log text" block, I can press `Enter` and trigger the callback that calls this function. But I can't find any way to actually change the value, so the check below also makes the function return early.

I suppose it's harmless to have this, and maybe there's some case I've overlooked which means it's actually necessary.

addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
```diff
@@ -56,6 +56,22 @@ var _drag_start: Vector2 = Vector2.INF
...
+func _set_editable(value) -> void:
+	if not is_node_ready():
+		await ready
+	editable = value
+	_line_edit.editable = value
+	_x_line_edit.editable = value
+	_y_line_edit.editable = value
+	_v3_x_line_edit.editable = value
+	_v3_y_line_edit.editable = value
+	_v3_z_line_edit.editable = value
```
wjt, line 73: You could do this:

```suggestion
	for line_edit in _last_submitted_text:
		line_edit.editable = value
```

though I don't know if it's an accident that these 6 controls happen to match that dictionary.

addons/block_code/ui/blocks/utilities/parameter_input/parameter_input.gd
manuq: @wjt you are right, this isn't necessary at all. I added this early return to avoid the emition of the `modified` signal at the end of this function. Nothing should be connected to that signal anyways for blocks in the picker. So do you think is better to remove the early returns from this patch?

wjt: I think it's okay to keep it, I was just wondering what it's here for.

wjt: (I think it is weird that signals like `LineEdit.text_submitted` can be emitted even when the field is not editable…)

manuq: Yep, I think the same.

Issue #356, opened 2025-01-23
A CodeBlock should not be allowed to attach to a node that has a script attached already


tindrew:
### What Happened

I discovered this purely by accident, but it should highlight some new functionality that needs to be added.  

If there's a script on a node, you can still add a BlockCode to it.  I assume that is not behavior that should happen, especially since it causes problems. 

![Image](https://github.com/user-attachments/assets/986edbbf-391f-4fd0-9f6c-9ec9e9c3cc50)

If I had to pick a fix for it, I'd say if you to attach a BlockCode to a node that has a script already, you should get a popup saying "You can't add a BlockCode to a node with an existing script!" (and also not let you add it of course).

### How to Reproduce

1. Add any node to a scene
2.  Add a script such as the starter player script
3. Add a block to the same node

### Block Coding Plugin Version

v0.9.0

### Godot Engine Version

v4.3.0 Stable

### Operating System

Windows 10

### Logs or command-line output

```shell

```

wjt:
I think that the warning that was added in #361 is too eager to fire. Take a look at the pong example:

![Image](https://github.com/user-attachments/assets/83199124-c86a-4c70-965d-730c7510e9ea)

The warning fires for both of the paddles because they have a script attached (namely `simple_character.gd`) and also a `BlockCode` node as a child. However the warning is false: the `BlockCode` node is smart enough to generate code that `extends SimpleCharacter` so there is no problem here.

manuq:
Oh good catch @wjt, my fault, thanks for noting this.

PR #370, opened 2025-01-29
Don't warn when BlockCode script extends parent script

wjt:
Commit 87c31b05c9f720cc2512285340117b86afc94e8e ("Warn when Block Code overrides parent script") introduced a warning when a BlockCode node's parent has a script attached. But if the parent is (for example) a Character2D with our SimpleCharacter script attached, this warning is incorrect: the block code script will correctly extend, not override, the parent.

Before issuing this warning, check if the parent's script has a class_name that matches what the generated BlockCode script will extend, and suppress the warning if so.

Fixes https://github.com/endlessm/godot-block-coding/issues/356

### DIFF
--- a/addons/block_code/block_code_node/block_code.gd
+++ b/addons/block_code/block_code_node/block_code.gd
@@ -85,7 +85,9 @@ func _get_configuration_warnings():
 	if get_parent() is BlockCode:
 		warnings.append("The parent must not be a BlockCode.")
 	if get_parent().script:
-		warnings.append("This BlockCode will override the existing script in the parent node.")
+		var parent_script_name: StringName = get_parent().script.get_global_name()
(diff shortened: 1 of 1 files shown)


wjt:
Before:

TODO

After:

image

addons/block_code/block_code_node/block_code.gd
```diff
@@ -85,7 +85,9 @@ func _get_configuration_warnings():
 	if get_parent() is BlockCode:
 		warnings.append("The parent must not be a BlockCode.")
 	if get_parent().script:
-		warnings.append("This BlockCode will override the existing script in the parent node.")
+		var parent_script_name: StringName = get_parent().script.get_global_name()
+		if not (parent_script_name and block_script and parent_script_name == block_script.script_inherits):
```
manuq, line 89: :+1:

manuq (approved):

Issue #365, opened 2025-01-28
Document nodes providing high-level-abstraction blocks


upost:
This allows games to create dynamic nodes, i.e. asteroids moving from right to left (placing one "original" asteroid outside of the screen, with on_timer_timeout and a random condition duplicate it and add it to a group, and then move all nodes in that group to the left until leaving the screen, then remove the node)

manuq:
@upost we have a SimpleSpawner node for precisely this behavior! See the talk here where it is used: https://www.youtube.com/watch?v=EJDNYFha9GY&t=523s . Or see it in action in the demos (also available through the Godot Asset Store): https://github.com/endlessm/godot-block-coding-demos

upost:
I didn't know that one, maybe I'm just the guy looking for docs and not youtube talks ;-)

manuq:
You are right @upost, we need documentation. Let's reopen this for documenting the SimpleSpawner.

PR #371, opened 2025-01-29
High level docs

manuq:
Fix https://github.com/endlessm/godot-block-coding/issues/365


manuq:
I started writting docs for users, but also ended up updating and cleaning the outdated info for devs too.

Issue #373, opened 2025-01-30
Add mouse position block


manuq:
### What Happened

This is a frequent request.

### How to Reproduce

N/A, it's a feature.

### Logs or command-line output

```shell

```

PR #374, opened 2025-01-30
Add new value block: mouse_position

manuq:
Under category Input. Considering the viewport scale as explained in [1]. The same thing we do for the viewport_width and other blocks.

[1] https://docs.godotengine.org/en/stable/tutorials/inputs/mouse_and_input_coordinates.html

Fix https://github.com/endlessm/godot-block-coding/issues/373


addons/block_code/blocks/input/mouse_position.tres
```diff
@@ -0,0 +1,21 @@
...
+description = "The coordinates of the mouse position."
+category = "Input"
+type = 3
+variant_type = 5
+display_template = "mouse position"
+code_template = "(func (): var transform: Transform2D = get_viewport_transform(); var scale: Vector2 = transform.get_scale(); return
+get_viewport().get_mouse_position() * scale ).call()
+
+
+"
```
wjt, line 18: Why not write this as:

```suggestion
code_template = "get_viewport().get_mouse_position() * get_viewport_transport().get_scale()"
```

I tried the following block program for the left paddle in pong:

image

The generated script is:

```gdscript
extends SimpleCharacter


func _process(delta):
	position = ((func (): var transform: Transform2D = get_viewport_transform(); var scale: Vector2 = transform.get_scale(); return
	get_viewport().get_mouse_position() * scale ).call()


	)

```

And this fails to start with:

> Parser Error: Expected closing ")" after grouping expression.

If I delete all the trailing newlines in this template, the generated code becomes:

```gdscript
extends SimpleCharacter

func _process(delta):
	position = ((func (): var transform: Transform2D = get_viewport_transform(); var scale: Vector2 = transform.get_scale(); return
	get_viewport().get_mouse_position() * scale ).call())

```

but the error is the same. Maybe it's the newline after `return`, how did that get in there...

Godot itself keeps crashing when I try to save/run changes... 

When I change the template to what I proposed above, the game runs, but the scaling is wrong. It could be that my program is wrong...

wjt: [Screencast from 2025-01-30 18-16-11.webm](https://github.com/user-attachments/assets/e3d42f90-e312-41ea-84e8-6acd950c6196)

wjt (changes requested):

addons/block_code/blocks/input/mouse_position.tres
manuq: Maybe we need to add `- get_viewport_transform().origin`? Thanks for reminding me that Pong is the perfect way to test this, as it has the viewport scaled.

wjt (approved):

PR #381, opened 2025-02-07
Split Script view that updates generated code as blocks are added/ removed

TakunHiwatari:
- Script Window code moved from block_code_plugin.gd to main_panel.gd
- Added Split Script scene
- "Show Generated Code" button now toggles Split Script view
- Split Script view has an "Open Window" button that opens the Script Window
- Split Script view becomes hidden when Script Window is open; users can no longer open multiple Script Windows
- Split Script and Script Window update generated code as blocks are added/ removed

https://github.com/user-attachments/assets/02c9fc96-f6f7-45ff-89ae-bb5a45733e3c


FocusedPhoenixGames:
I'm from the ReBlocks team and I've been working on improving the Generated Script views. These changes allow flexibility for users to see the generated code automatically update and they have the option of viewing it in the editor, or use the window for a 2nd monitor.

I've also added a HSplitContainer to the Main Scene so the split script window is now resizable. The video below demonstrates this and the generated code updating when new blocks are added and removed.

https://github.com/user-attachments/assets/3343fb34-9b40-48cc-ba59-edd378c0a68e

tindrew (approved):

PR #393, opened 2025-06-05
Localized properties

manuq:
[Grabación de pantalla desde 2025-06-05 16-00-33.webm](https://github.com/user-attachments/assets/ce224afc-7a40-411b-b35e-57900dcd2522)

### DIFF
--- a/addons/block_code/examples/pong_game/player_score.tscn
+++ b/addons/block_code/examples/pong_game/player_score.tscn
@@ -1,11 +1,11 @@
 [gd_scene load_steps=11 format=3 uid="uid://djmtbm15n2wqq"]
 
-[ext_resource type="Script" path="res://addons/block_code/simple_nodes/simple_scoring/simple_scoring.gd" id="1_eafo0"]
-[ext_resource type="Script" path="res://addons/block_code/block_code_node/block_code.gd" id="2_4mu48"]
-[ext_resource type="Script" path="res://addons/block_code/serialization/block_serialization_tree.gd" id="3_kpcgt"]
-[ext_resource type="Script" path="res://addons/block_code/serialization/block_serialization.gd" id="4_ph8ne"]
-[ext_resource type="Script" path="res://addons/block_code/serialization/block_script_serialization.gd" id="6_k7up7"]
-[ext_resource type="Script" path="res://addons/block_code/code_generation/variable_definition.gd" id="7_x0rr3"]
+[ext_resource type="Script" uid="uid://chi1008f4c7el" path="res://addons/block_code/simple_nodes/simple_scoring/simple_scoring.gd" id="1_eafo0"]
+[ext_resource type="Script" uid="uid://cumxsbxj56dmk" path="res://addons/block_code/block_code_node/block_code.gd" id="2_4mu48"]
+[ext_resource type="Script" uid="uid://brx1cabdmgu38" path="res://addons/block_code/serialization/block_serialization_tree.gd" id="3_kpcgt"]
+[ext_resource type="Script" uid="uid://u035wsjd5tpj" path="res://addons/block_code/serialization/block_serialization.gd" id="4_ph8ne"]
+[ext_resource type="Script" uid="uid://tlsvrkc3fubl" path="res://addons/block_code/serialization/block_script_serialization.gd" id="6_k7up7"]
+[ext_resource type="Script" uid="uid://bwkinr8omutpw" path="res://addons/block_code/code_generation/variable_definition.gd" id="7_x0rr3"]
 
 [sub_resource type="Resource" id="Resource_paaek"]
 script = ExtResource("4_ph8ne")
@@ -44,7 +44,6 @@ version = 0
 
 [node name="PlayerScore" type="Node2D" groups=["hud"]]
 script = ExtResource("1_eafo0")
-score = null
 
 [node name="BlockCode" type="Node" parent="."]
 script = ExtResource("2_4mu48")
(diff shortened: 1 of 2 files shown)


manuq:
The last commit makes the setter and changer blocks translatable:
Captura desde 2025-06-06 14-42-16

addons/block_code/ui/blocks/block/block.gd
```diff
@@ -129,6 +129,13 @@ func _get_format_string() -> String:
...
 		return ""
 
+	if definition.property_name:
+		var domain: TranslationDomain = TranslationServer.get_or_add_domain("godot.properties")
+		var translated_property: String = domain.translate(definition.property_name.capitalize())
+		# TODO: Ideally we should be also passing the context. See:
+		# https://github.com/godotengine/godot/blob/978b38797ba8e8757592f21101e32e364d60662d/editor/editor_property_name_processor.cpp#L90
+		return tr(definition.display_template) % translated_property.to_lower()
+
 	return tr(definition.display_template)
```
wjt, line 139: I might add a note to the `display_template` docstring to the effect that

```gdscript
## If [member property_name] is set, this template is assumed to be a format string with a `%s` placeholder; in this case, any literal `%` signs must be escaped as `%%`.
```

You could imagine this being a problem because `%` is a unit that a property might have. (I think it's fine to do it this way.)

addons/block_code/ui/block_canvas/block_canvas.gd
```diff
@@ -133,21 +133,35 @@ func _drop_node(at_position: Vector2, data: Variant) -> void:
...
+static func _get_classname_for_property(_class_name: StringName, property_name: StringName) -> StringName:
+	var return_classname := &""
+	for parent_class_name in BlocksCatalog.get_parents(_class_name):
+		var property_list := ClassDB.class_get_property_list(parent_class_name, false)
+		var has_property = property_list.any(func(dict): return dict.name == property_name)
+		if has_property:
+			return_classname = parent_class_name
+			continue
+		else:
+			break
```
wjt, line 150: Are the `continue` and `break` here the wrong way around?

I think what you're trying to do here is to walk up the ancestor classes and return the first class that defines a property of that name. If D is a subclass of C is a subclass of B is a subclass of A, and B and A both define property `foo` then `_get_classname_for_property("C", "foo")` should return `"B"`, as should `_get_classname_for_property("D", "foo")`.

If that's correct then I believe this function is wrong:

- `_get_classname_for_property("C", "foo")` will return `"A"`
- `_get_classname_for_property("D", "foo")` will return `""` because it breaks out of the loop at the first class that does *not* have the property

Assuming I'm right here, you can swap `continue` and `break` and the function will behave correctly.

Personally I find the control flow easier to read by `return`ing from the body of the loop, rather than having a `var return_classname` and `return`ing only at the end of the function:

```gdscript
static func _get_classname_for_property(_class_name: StringName, property_name: StringName) -> StringName:
	for parent_class_name in BlocksCatalog.get_parents(_class_name):
		var property_list := ClassDB.class_get_property_list(parent_class_name, false)
		var has_property = property_list.any(func(dict): return dict.name == property_name)
		if has_property:
			return parent_class_name

	return &""
```

addons/block_code/code_generation/block_definition.gd
```diff
@@ -215,11 +215,11 @@ static func new_property_setter(_class_name: String, property: Dictionary, categ
 	var block_definition: Resource = new(
 		PROPERTY_SETTER_NAME_FORMAT % [_class_name, property.name],
 		_class_name,
-		"Set the %s property" % property.name,
+		Engine.tr("Set the %s property") % property.name,
```
wjt, line 221: TIL. ¡Gracias, Godot Meetup Argentina!

wjt (changes requested):
Nice work! I spotted what looks like a logic error but I could be wrong – I haven't run the code to check.

addons/block_code/ui/blocks/block/block.gd
manuq: Oh good point. It is a bit weird that this PR adds 2 paths for `display_template`, one that formats the translated string and one that doesn't. At least let's document it.

addons/block_code/ui/block_canvas/block_canvas.gd
wjt: Oh! I completely missed the `false` parameter. I don't like the design of that API: boolean parameter with negated sense!

addons/block_code/code_generation/block_definition.gd
```diff
@@ -2,8 +2,17 @@
...
 const Types = preload("res://addons/block_code/types/types.gd")
+const VariableDefinition = preload("res://addons/block_code/code_generation/variable_definition.gd")
 
 const FORMAT_STRING_PATTERN = "\\[(?<out_parameter>[^\\]]+)\\]|\\{const (?<const_parameter>[^}]+)\\}|\\{(?!const )(?<in_parameter>[^}]+)\\}|(?<label>[^\\{\\[]+)"
+const PROPERTY_SETTER_NAME_PATTERN = "(?<class_name>[^\\s]*)_property_set_(?<property_name>[^\\s]+)"
+const PROPERTY_SETTER_NAME_FORMAT = &"%s_property_set_%s"
+const PROPERTY_CHANGER_NAME_PATTERN = "(?<class_name>[^\\s]*)_property_change_(?<property_name>[^\\s]+)"
+const PROPERTY_CHANGER_NAME_FORMAT = &"%s_property_change_%s"
+const PROPERTY_GETTER_NAME_PATTERN = "(?<class_name>[^\\s]*)_property_get_(?<property_name>[^\\s]+)"
+const PROPERTY_GETTER_NAME_FORMAT = &"%s_property_get_%s"
```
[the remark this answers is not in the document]
manuq: So yes, I managed to make the PR not having any breaking changes, and rebased.

Issue #400, opened 2025-07-12
Problem with variables in Godot 4.4


KaiSD:
### What Happened

I was trying to use Block Coding in Godot 4.4 Stable.

As a test, I created a 2d node, created a variable at start and tried to access it in "every frame".

For all kinds of variables I'm getting the same error, that looks like this:
1. For trying to use numerical variable (float or int) in calculation or condition: "Parser Error: Invalid operands "Node" and "float" for "-" operator."
2. For trying to getvalue from Vector2: "Invalid access to property or key 'y' on a base object of type 'null instance'."

I looked at the generated code, and seen that variable's name in the gnerated code in soce situations left as "(%s)".

<img width="1289" height="922" alt="Image" src="https://github.com/user-attachments/assets/4c1174d6-8e61-4f96-9dee-6e788ea58dd9" />

### How to Reproduce

1. Set the variable when starting
2. Try "variable = variable - 1" or "variable = variable + 1" in every frame block.
3. In resulting code some variable names will be left as (%s), and the code will not work.

### Block Coding Plugin Version

v0.10.0

### Godot Engine Version

v4.4 Stable

### Logs or command-line output

```shell

```

bront32:
I hope this can be fixed soon, I was hoping to transition school students from Scratch to Godot using Block Coding and the Vector variables are essential.  I wish I knew how to look into it myself but could not follow the scripts.

PR #402, opened 2025-08-06
Repopulate variable code template

DoomTas3r:
Looks like a simple writing mistake!

Fix for https://github.com/endlessm/godot-block-coding/issues/400
Caused by [947bd28](https://github.com/endlessm/godot-block-coding/commit/947bd28888551a91fff86ee0ad0506eb27335109)


manuq (approved):
Indeed, thanks for your contribution! It is very nice that you also referenced the exact commit with the regression, the error is clear in the commit diff.

Issue #360, opened 2025-01-24
Variables: Allow deletion


manuq:
### What Happened

Variables can be created from the Variables section using the "+ Create New Variable" button. But they can't be removed. Add UI below the "+ Create New Variable" button to remove existing variables. This could be a "🗑 Remove Variable" dropdown, with a confirmation dialogue.

### How to Reproduce

N/A

### Block Coding Plugin Version

main

### Godot Engine Version

4.3

### Operating System

Endless OS

### Logs or command-line output

```shell

```

tindrew:
I reported this to you over 7 months ago. Glad to see you finally fixed it.

PR #403, opened 2025-08-07
Adds delete variable button and dialogue

DoomTas3r:
Fix for #360

Adds the delete variable button with a dialogue box and check boxes to select variable names and delete them if they exist. The delete variable functionality resembles the create variable functionality, including the ability to undo.

Functionality missing from this commit might include hiding the delete variables button when there are no variables, and disabling the delete button if no check boxes are checked.

The missing title bar, the outline, and the curved corners are from my window manager.
dialogue

button


manuq (approved):
Thank you!

PR #404, opened 2025-08-08
Adds to_string block with optional types

DoomTas3r:
Partially addresses #351

The to_string block converts a parameter to a string. It includes an option input with an integer, a boolean, a vector2, a vector3, a color, and an object.

The type of the selected input gets detected and displays as a normal parameter input with the detected type and snapping
example

This prints:
```
4.0
false
(0.0, 0.0)
(1.0, 1.0, 1.0, 1.0)
Sprite2D:<Sprite2D#59659781447>
```


manuq (approved):
Yes I think is good to have this block for now. Thank you!

